PR #13, opened 2017-01-27
Allow iam roles to be admins

fieldju:
Add code so that CMS can be configured to allow instances to have admin rights.
This is needed to enable the the cross region backup lambda to have access to the metadata admin endpoint.

### DIFF
--- a/src/main/java/com/nike/cerberus/security/VaultAuthPrincipal.java
+++ b/src/main/java/com/nike/cerberus/security/VaultAuthPrincipal.java
@@ -66,11 +66,7 @@ public class VaultAuthPrincipal implements Principal {
         final ImmutableSet.Builder<String> roleSetBuilder = ImmutableSet.builder();
         final Map<String, String> meta = clientToken.getMeta();
 
-        if (meta == null || meta.isEmpty()) {
-            throw new IllegalArgumentException("client token does not contain expected metadata!");
-        }
-
-        if (Boolean.valueOf(meta.get(METADATA_KEY_IS_ADMIN))) {
+        if (meta != null && Boolean.valueOf(meta.get(METADATA_KEY_IS_ADMIN))) {
             roleSetBuilder.add(ROLE_ADMIN);
         }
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/nike/cerberus/service/AuthenticationService.java
```diff
@@ -387,4 +405,17 @@ private String getKeyId(IamRoleCredentials credentials) {
...
         }
     }
+
+    private Set<String> getAdminRoleArnSet() {
+        if (adminRoleArnSet == null) {
+            adminRoleArnSet = new HashSet<>();
+            if (StringUtils.isNotBlank(adminRoleArns)) {
+                String[] roles = adminRoleArns.split(",");
+                if (roles.length > 0) {
+                    Arrays.stream(roles).forEach(adminRoleArnSet::add);
```
fieldju: - [ ] Trim the white space so `foo,bar` and `foo, bar` are valid

src/main/java/com/nike/cerberus/security/VaultAuthPrincipal.java
```diff
@@ -81,7 +77,7 @@ public VaultAuthPrincipal(VaultClientTokenResponse clientToken) {
 
     private Set<String> extractUserGroups(final VaultClientTokenResponse clientToken) {
         final Map<String, String> meta = clientToken.getMeta();
-        final String groupString = meta.get(METADATA_KEY_GROUPS);
+        final String groupString = meta == null ? "" : meta.get(METADATA_KEY_GROUPS);
```
tlisonbee: Don't need the null check here, StringUtils.isBlank() handles doesn't it?

src/main/java/com/nike/cerberus/security/VaultAuthPrincipal.java
```diff
@@ -91,7 +87,7 @@ public VaultAuthPrincipal(VaultClientTokenResponse clientToken) {
 
     private String extractUsername(final VaultClientTokenResponse clientToken) {
         final Map<String, String> meta = clientToken.getMeta();
-        return meta.get(METADATA_KEY_USERNAME);
+        return meta == null ? "not-set-possibly-root-token" : meta.get(METADATA_KEY_USERNAME);
```
tlisonbee: This line is awkward.  It probably at least deserves a comment.

Is there better wording for "not-set-possibly-root-token"?

src/main/java/com/nike/cerberus/security/VaultAuthPrincipal.java
fieldju: The object can be null, so if it is we want it to be empty, or we want to get the groups from the metadata.

src/main/java/com/nike/cerberus/security/VaultAuthPrincipal.java
fieldju: - [ ] Add comment, use better username like root, sense it can only happen with root token, or tokens created manually with the vault api outside of CMS.

PR #14, opened 2017-01-27
Add an Okta connector that always requires MFA

sdford:
Adds a new Okta connector that requires MFA for Cerberus login, regardless of whether the user's account requires it.


src/main/java/com/nike/cerberus/auth/connector/okta/OktaAuthHelper.java
```diff
@@ -154,12 +171,130 @@ protected AuthResult authenticateUser(final String username, final String passwo
...
+     * Get username from authentication result
+     * @param authResult Okta response object after successful authentication
+     * @return The username
+     */
+    String getUserLoginFromAuthResult(final AuthResult authResult) {
+        final EmbeddedAuthResponseDataV1 embeddedAuthData = this.getEmbeddedAuthData(authResult);
+
+        try {
+            return embeddedAuthData.getUser().getProfile().getLogin();
+        } catch(NullPointerException npe) {
```
tlisonbee: Catching NPE here seems slightly off to me.  Is there a more specific check for "Could not parse user data from Okta response"?  NPE is so general, it might mask other bugs.

src/main/java/com/nike/cerberus/auth/connector/okta/OktaAuthHelper.java
```diff
@@ -154,12 +171,130 @@ protected AuthResult authenticateUser(final String username, final String passwo
...
+    }
+
+    /**
+     * Ensure the user has at least one active MFA device set up.
+     * @param factors - List of user factors
+     */
+    protected void validateUserFactors(final List<Factor> factors) {
+
+       if(factors == null || factors.isEmpty() || factors.stream()
+               .allMatch(factor -> StringUtils.equals(factor.getStatus(), MFA_FACTOR_NOT_SETUP_STATUS)))
```
tlisonbee: Is the negative check right?  Seems like it might be better to do a positive check that "at least one MFA Factor is setup"?

src/main/java/com/nike/cerberus/auth/connector/okta/OktaMFAAuthConnector.java
```diff
@@ -0,0 +1,112 @@
...
+        final AuthResult authResult = oktaAuthHelper.authenticateUser(username, password, null);
+        final String userId = this.oktaAuthHelper.getUserIdFromAuthResult(authResult);
+        final String userLogin = this.oktaAuthHelper.getUserLoginFromAuthResult(authResult);
+
+        final AuthData authData = new AuthData().setUserId(userId).setUsername(userLogin);
+        final AuthResponse authResponse = new AuthResponse().setData(authData).setStatus(AuthStatus.MFA_REQUIRED);
+
+        final List<Factor> factors;
+
+        if (StringUtils.equals(authResult.getStatus(), OktaAuthHelper.AUTHENTICATION_MFA_REQUIRED_STATUS)
```
tlisonbee: Nitpick: Rather than StringUtils.equals( ..., ...) this would read nicer if it was encapsulated in a class.

E.g. oktaAuthStatus.isMfaRequired(authResult.getStatus()) or similar

src/main/java/com/nike/cerberus/auth/connector/okta/OktaAuthHelper.java
sdford: I agree, but Okta's API unfortunately does not alway provide a status on the factors. It seems that the only status that is guaranteed is the negative ("NOT_SETUP").

tlisonbee: Got it, makes sense. Thanks.

PR #21, opened 2017-02-21
Refactor get metadata to use services rather than DAOs directly and a…

fieldju:
…dd new endpoint for restoring metadata

### DIFF
--- a/src/main/java/com/nike/cerberus/domain/SafeDepositBox.java
+++ b/src/main/java/com/nike/cerberus/domain/SafeDepositBox.java
@@ -164,4 +164,46 @@ public class SafeDepositBox {
     public void setIamRolePermissions(Set<IamRolePermission> iamRolePermissions) {
         this.iamRolePermissions = iamRolePermissions;
     }
+
+    @Override
+    public boolean equals(Object o) {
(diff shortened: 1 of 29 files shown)


src/main/java/com/nike/cerberus/domain/SafeDepositBox.java
```diff
@@ -164,4 +164,46 @@ public void setUserGroupPermissions(Set<UserGroupPermission> userGroupPermission
...
     }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        SafeDepositBox that = (SafeDepositBox) o;
+
+        if (id != null ? !id.equals(that.id) : that.id != null) return false;
```
sdford: What do you think of making these statements positive so that they're easier to read?
e.g. `( ! ( id == null ? that.id == null : id.equals(that.id) ) ) return false;`

sdford: Or having a method for this since all of these 'if' statements seem to be doing the same thing

PR #25, opened 2017-03-29
Validate KMS policies at IAM authentication time and recreate polcies…

fieldju:
… where a consumer has deleted and recreated an IAM role causing the KMS key policy to reference an ID rather than the user supplied ARN

### DIFF
--- a/src/main/java/com/nike/cerberus/service/KmsService.java
+++ b/src/main/java/com/nike/cerberus/service/KmsService.java
@@ -32,6 +32,8 @@ import com.nike.cerberus.error.DefaultApiError;
 import com.nike.cerberus.record.AwsIamRoleKmsKeyRecord;
 import com.nike.cerberus.util.UuidSupplier;
 import org.mybatis.guice.transactional.Transactional;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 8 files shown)


src/main/java/com/nike/cerberus/service/KmsService.java
```diff
@@ -105,4 +111,46 @@ public String provisionKmsKey(final String iamRoleId,
...
+                    .withApiErrors(DefaultApiError.FAILED_TO_VALIDATE_KMS_KEY_POLICY)
+                    .withExceptionCause(e)
+                    .withExceptionMessage(
+                            String.format("Failed to validate KMS key policy for keyId: " +
+                                    "%s for IAM role: %s in region: %s", keyId, iamRoleArn, kmsCMKRegion))
+                    .build();
+        }
+
+        if (! kmsPolicyService.isPolicyValid(policyResult.getPolicy(), iamRoleArn)) {
+            String updatedPolicy = kmsPolicyService.generateStandardKmsPolicy(iamRoleArn);
```
tlisonbee: Maybe we should log info() here or above when we find invalid ones?  It isn't going to happen that much and might be helpful.

fieldju: Done

PR #26, opened 2017-03-29
Store IAM role ARNs in the DB

sdford:
Migrate DB to store role ARNs and use ARNs for role lookup. Account id and role name are still stored for now.


src/test/java/com/nike/cerberus/util/AwsIamRoleArnParserTest.java
```diff
@@ -0,0 +1,73 @@
...
+    private AwsIamRoleArnParser awsIamRoleArnParser;
+
+    @Before
+    public void setup() {
+
+        awsIamRoleArnParser = new AwsIamRoleArnParser();
+    }
+
+    @Test(expected = RuntimeException.class)
+    public void getArnFailsNullRoleName() {
```
tlisonbee: Naming convention for test code in this project should use underscores.

src/main/java/com/nike/cerberus/util/AwsIamRoleArnParser.java
```diff
@@ -0,0 +1,50 @@
...
+
+    public static final String AWS_IAM_ROLE_ARN_TEMPLATE = "arn:aws:iam::%s:role/%s";
+
+    private static final Pattern IAM_ROLE_ARN_PATTERN = Pattern.compile("^arn:aws:iam::(?<accountId>\\d+?):role/(?<roleName>.+)$");
+
+    public String getAccountId(String roleArn) {
+
+        Matcher iamRoleArnMatcher = IAM_ROLE_ARN_PATTERN.matcher(roleArn);
+
+        if (! iamRoleArnMatcher.find()) {throw new RuntimeException("Invalid role ARN");}
```
tlisonbee: Instead of RuntimeException how about define your own Exception type that extends RuntimeException?

tlisonbee: Wait, this is temporary code, it doesn't matter

src/test/java/com/nike/cerberus/util/AwsIamRoleArnParserTest.java
tlisonbee: Does this method have bad method name too?  isn't this a null account id test?

PR #28, opened 2017-03-31
Add a V2 authentication endpoint for IAM roles

sdford:
Adds a v2/auth/iam-role endpoint that only accepts IAM role ARNs, instead of account id and role name.

The v1 endpoint still exists, but has been deprecated.


fieldju:
You need to document the new endpoint in the API.md

PR #30, opened 2017-04-03
Feature/add principal event logging

fieldju:


src/main/java/com/nike/cerberus/endpoints/sdb/GetSafeDepositBox.java
```diff
@@ -62,10 +66,16 @@ public GetSafeDepositBox(final SafeDepositBoxService safeDepositBoxService) {
 
         if (securityContext.isPresent()) {
             final VaultAuthPrincipal vaultAuthPrincipal = (VaultAuthPrincipal) securityContext.get().getUserPrincipal();
+
+            String sdbId = request.getPathParam("id");
+            Optional<String> sdbNameOptional = safeDepositBoxService.getSafeDepositBoxNameById(sdbId);
+            String sdbName = sdbNameOptional.isPresent() ? sdbNameOptional.get() : String.format("(Failed to lookup name from id: %s)", sdbId);
+            log.info("Read SDB Event: the principal: {} is attempting to read sdb name: '{}'", vaultAuthPrincipal.getName(), sdbName);
```
tlisonbee: Should we log SDB id too since we have it?

For example, I'm not sure if the ID would help us correlate with other logging or similar.

PR #33, opened 2017-04-10
Add V2 Safe Deposit Box API with ARNs

sdford:
* Adds a v2/safe-deposit-box API that uses IAM principal ARNs instead of account id, and role name
* Updates v2/auth/iam-role to accept 'iam_principal_arn' instead of 'role_arn'

- [x]  Return safe deposit box after POST and PUT
- [x]  Remove unused account id and role name columns


tlisonbee:
It seems like the drop columns should be part of this pull request too.  Does that seem right?

sdford:
@tlisonbee I think that makes sense; added.

PR #36, opened 2017-04-27
Avoid hitting API limit on KMS policy validation

sdford:
* Limit KMS policy validation for each KMS key to once per minute (at most) by default


README.md
```diff
@@ -59,19 +59,20 @@ That will setup the default policy and generate a token for CMS and output:
...
+cms.role.arn                                        | Yes | The arn for the Instance profile for CMS instances, so they can admin KMS keys that they create.
+cms.admin.group                                     | Yes | Group that user can be identified by to get admin privileges, currently this just enables users to access `/v1/metadata` see API.md
+cms.admin.roles                                     | No  | Comma seperated list of ARNs that can auth and access admin endpoints.
+cms.auth.connector                                  | Yes | The user authentication connector implementation to use for user auth.
+cms.user.token.ttl.override                         | No  | By default user tokens are created with a TTL of 1h, you can override that with this param
+cms.iam.token.ttl.override                          | No  | By default IAM tokens are created with a TTL of 1h, you can override that with this param
+cms.kms.policy.validation.interval.millis.override  | No  | By default CMS validates KMS key policies no more than once per minute, you can override that with this param
```
fieldju: A little more context here will confuse people less, maybe something like.

KMS Policies are bound to IAM Principal IDs rather than ARNs themselves and because of this we validate the policy at auth time, to ensure if an IAM role has been deleted and re-created that we grant access to the new id. This API has a low limit, therefore we throttle how often we check ......

something along those lines.

Issue #18, opened 2017-02-04
When a iam role authenticates with an invalid region, a 500 is returned instead of a 400


fieldju:
```
traceId=e7e8ba519c741e85 2017-02-04T05:08:44,955Z [epollEventLoopGroup-3-7] appname=cms environment=cerberus-test version=0.12.0 |-ERROR c.n.b.h.r.RiposteUnhandledExceptionHandler - Caught unhandled exception: error_uid=e3dc74a9-afba-48f4-a1e9-2494e23b89e5, dtrace_id=e7e8ba519c741e85, exception_class=java.util.concurrent.CompletionException, returned_http_status_code=500, contributing_errors="GENERIC_SERVICE_ERROR", request_uri="/v1/auth/iam-role", request_method="POST", query_string="null", request_headers="Accept=application/json,CloudFront-Viewer-Country=US,CloudFront-Forwarded-Proto=https,CloudFront-Is-Tablet-Viewer=false,CloudFront-Is-Mobile-Viewer=false,X-Forwarded-Proto=https,Connection=keep-alive,CloudFront-Is-SmartTV-Viewer=false,X-Forwarded-Port=443,Via=1.1 a970721b7e7e7717531fc56b64b3884f.cloudfront.net (CloudFront),host=xxxxxxxxxxxxxx,X-Amz-Cf-Id=VPUCT40x9KqdIQV4fwZN4Pjv1oYYYER4RrIqh6jjyJFBFlIAiVGtLg==,X-Forwarded-For=52.55.130.29, 54.239.145.100, 172.20.0.252, 172.20.8.39,content-type=application/json,Content-Length=88,X-Real-IP=172.20.0.252,CloudFront-Is-Desktop-Viewer=true", unhandled_error="true"
java.util.concurrent.CompletionException: java.lang.IllegalArgumentException: Specified region is not valid.
	at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:273)
	at java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:280)
	at java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1592)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
[...]

PR #37, opened 2017-05-05
Fix incorrect error codes, and add better messaging

sdford:
* Return the proper error codes and messages for:
    * Bad username, password or MFA
    * Invalid region during IAM authentication (Issue #18)
    * Duplicate SDB names

### DIFF
--- a/src/main/java/com/nike/cerberus/auth/connector/onelogin/OneLoginAuthConnector.java
+++ b/src/main/java/com/nike/cerberus/auth/connector/onelogin/OneLoginAuthConnector.java
@@ -194,7 +194,7 @@ public class OneLoginAuthConnector implements AuthConnector {
                         .withApiErrors(DefaultApiError.MFA_SETUP_REQUIRED)
                         .withExceptionMessage(msg)
                         .build();
-            } else if (statusCode == 401) {
+            } else if (statusCode == 400 || statusCode == 401) {
                 throw ApiException.newBuilder()
                         .withApiErrors(DefaultApiError.AUTH_BAD_CREDENTIALS)
                         .withExceptionMessage(msg)
--- a/src/main/java/com/nike/cerberus/auth/connector/onelogin/OneLoginAuthConnector.java
+++ b/src/main/java/com/nike/cerberus/auth/connector/onelogin/OneLoginAuthConnector.java
@@ -157,7 +157,7 @@ public class OneLoginAuthConnector implements AuthConnector {
             String msg = String.format("stateToken: %s failed to verify 2nd factor for reason: %s",
                     stateToken, verifyFactorResponse.getStatus().getMessage());
 
-            if (verifyFactorResponse.getStatus().getCode() == 400L) {
+            if (verifyFactorResponse.getStatus().getCode() == 401) {
                 throw ApiException.newBuilder()
                         .withApiErrors(DefaultApiError.AUTH_BAD_CREDENTIALS)
                         .withExceptionMessage(msg)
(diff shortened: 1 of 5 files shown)


src/main/java/com/nike/cerberus/auth/connector/onelogin/OneLoginAuthConnector.java
```diff
@@ -157,7 +157,7 @@ protected SessionLoginTokenData verifyFactor(final String deviceId,
             String msg = String.format("stateToken: %s failed to verify 2nd factor for reason: %s",
                     stateToken, verifyFactorResponse.getStatus().getMessage());
 
-            if (verifyFactorResponse.getStatus().getCode() == 400L) {
+            if (verifyFactorResponse.getStatus().getCode() == 401) {
```
fieldju: You are introducing a bug here.

This code is verifying the response code from the OneLogin API call, which would still be a `400`, unless I am mistaken.

src/main/java/com/nike/cerberus/auth/connector/onelogin/OneLoginAuthConnector.java
```diff
@@ -184,20 +184,21 @@ protected SessionLoginTokenData createSessionLoginToken(final String username, f
...
-                            .withApiErrors(DefaultApiError.MFA_SETUP_REQUIRED)
-                            .withExceptionMessage(msg)
-                            .build();
-                } else {
-                    throw ApiException.newBuilder()
-                            .withApiErrors(DefaultApiError.AUTH_BAD_CREDENTIALS)
-                            .withExceptionMessage(msg)
-                            .build();
-                }
+            if (statusCode == 400 &&
```
fieldju: Again, unless OneLogin changed there API, this code is going to be introducing bugs no?

src/main/java/com/nike/cerberus/auth/connector/onelogin/OneLoginAuthConnector.java
```diff
@@ -184,20 +184,21 @@ protected SessionLoginTokenData createSessionLoginToken(final String username, f
...
+                    StringUtils.startsWithIgnoreCase(createSessionLoginTokenResponse.getStatus().getMessage(), "MFA")) {
+                throw ApiException.newBuilder()
+                        .withApiErrors(DefaultApiError.MFA_SETUP_REQUIRED)
+                        .withExceptionMessage(msg)
+                        .build();
+            } else if (statusCode == 401) {
+                throw ApiException.newBuilder()
+                        .withApiErrors(DefaultApiError.AUTH_BAD_CREDENTIALS)
```
fieldju: In DefaultApiError the status code for this needs to change from 400 -> 401

src/main/java/com/nike/cerberus/auth/connector/onelogin/OneLoginAuthConnector.java
```diff
@@ -184,20 +184,21 @@ protected SessionLoginTokenData createSessionLoginToken(final String username, f
...
+                        .withExceptionMessage(msg)
+                        .build();
             } else {
                 throw ApiException.newBuilder()
                         .withApiErrors(DefaultApiError.GENERIC_BAD_REQUEST)
```
fieldju: This is also where that `The request will not be completed` message is getting generated from and needs to be cleaned up with a better message and a `401`

src/main/java/com/nike/cerberus/auth/connector/onelogin/OneLoginAuthConnector.java
sdford: Unfortunately, OneLogin changed what they're returning here without changing the API version.

src/main/java/com/nike/cerberus/auth/connector/onelogin/OneLoginAuthConnector.java
sdford: Makes sense. I will fix that as a part of this PR too.

PR #41, opened 2017-05-12
Enable all principals to authenticate with Cerberus

sdford:
This fixes IAM authentication for IAM assumed roles via the v2 API, and enables all other principals to authenticate

### DIFF
--- a/src/main/java/com/nike/cerberus/service/AuthenticationService.java
+++ b/src/main/java/com/nike/cerberus/service/AuthenticationService.java
@@ -401,20 +401,22 @@ public class AuthenticationService {
         return policies;
     }
 
+    /**
+     * Build a set of policies that is allowed for the specific IAM principal (e.g. an assumed user, or instance profile)
+     * as well as the generic role ARN (i.e. arn:aws:iam::1111111111:role/example)
+     * @param iamPrincipalArn - The given IAM principal ARN during authentication
+     * @return - List of all policies the given ARN has access to
+     */
     protected Set<String> buildCompleteSetOfPolicies(final String iamPrincipalArn) {
 
         final Set<String> allPolicies = buildPolicySet(iamPrincipalArn);
 
-        try {
-            if (! awsIamRoleArnParser.isRoleArn(iamPrincipalArn)) {
-                logger.debug("Could not find IAM role in principal format, trying 'role' format...");
-                final String iamPrincipalInRoleFormat = awsIamRoleArnParser.convertPrincipalArnToRoleArn(iamPrincipalArn);
+        if (! awsIamRoleArnParser.isRoleArn(iamPrincipalArn)) {
+            logger.debug("Could not find IAM role in principal format, trying 'role' format...");
+            final String iamPrincipalInRoleFormat = awsIamRoleArnParser.convertPrincipalArnToRoleArn(iamPrincipalArn);
 
-                final Set<String> additionalPolicies = buildPolicySet(iamPrincipalInRoleFormat);
-                allPolicies.addAll(additionalPolicies);
-            }
-        } catch (Exception e) {
-            logger.debug("Failed to convert principal ARN to role ARN");
+            final Set<String> additionalPolicies = buildPolicySet(iamPrincipalInRoleFormat);
+            allPolicies.addAll(additionalPolicies);
         }
 
         return allPolicies;
@@ -448,7 +450,6 @@ public class AuthenticationService {
      */
     protected String getKeyId(IamPrincipalCredentials credentials) {
         final String iamPrincipalArn = credentials.getIamPrincipalArn();
-        final String region = credentials.getRegion();
         final Optional<AwsIamRoleRecord> iamRole = findIamRoleAssociatedWithSdb(iamPrincipalArn);
 
         if (!iamRole.isPresent()) {
@@ -458,14 +459,14 @@ public class AuthenticationService {
                     .build();
         }
 
-        final Optional<AwsIamRoleKmsKeyRecord> kmsKey = awsIamRoleDao.getKmsKey(iamRole.get().getId(), region);
+        final Optional<AwsIamRoleKmsKeyRecord> kmsKey = awsIamRoleDao.getKmsKey(iamRole.get().getId(), credentials.getRegion());
 
         final String kmsKeyId;
         final AwsIamRoleKmsKeyRecord kmsKeyRecord;
         final OffsetDateTime now = dateTimeSupplier.get();
 
         if (!kmsKey.isPresent()) {
-            kmsKeyId = kmsService.provisionKmsKey(iamRole.get().getId(), iamPrincipalArn, region, SYSTEM_USER, now);
+            kmsKeyId = kmsService.provisionKmsKey(iamRole.get().getId(), iamPrincipalArn, credentials.getRegion(), SYSTEM_USER, now);
         } else {
             kmsKeyRecord = kmsKey.get();
             kmsKeyId = kmsKeyRecord.getAwsKmsKeyId();
@@ -526,6 +527,12 @@ public class AuthenticationService {
         return adminRoleArnSet;
     }
 
+    /**
+     * Generate map of Vault token metadata that is common to all principals
+     * @param iamPrincipalArn - The authenticating IAM principal ARN
+     * @param region - The AWS region
+     * @return - Map of token metadata
+     */
     protected Map<String, String> generateCommonVaultPrincipalAuthMetadata(final String iamPrincipalArn, final String region) {
         Map<String, String> metadata = Maps.newHashMap();
         metadata.put(VaultAuthPrincipal.METADATA_KEY_AWS_REGION, region);
@@ -546,20 +553,22 @@ public class AuthenticationService {
         return metadata;
     }
 
+    /**
+     * Search for the given IAM principal (e.g. instance profile, or assumed user), if not found, then search for the
+     * generic role ARN (i.e. arn:aws:iam::1111111111:role/example)
+     * @param iamPrincipalArn - The authenticating IAM principal ARN
+     * @return - The associated IAM role record
+     */
     protected Optional<AwsIamRoleRecord> findIamRoleAssociatedWithSdb(final String iamPrincipalArn) {
         Optional<AwsIamRoleRecord> iamRole = awsIamRoleDao.getIamRole(iamPrincipalArn);
 
-        try {
-            // if the arn is not already in 'role' format, and cannot be found,
-            // then try checking for the generic "arn:aws:iam::0000000000:role/foo" format
-            if ( !( awsIamRoleArnParser.isRoleArn(iamPrincipalArn) || iamRole.isPresent() ) ) {
-                logger.debug("Could not find IAM role in principal format, trying 'role' format...");
-                final String iamPrincipalInRoleFormat = awsIamRoleArnParser.convertPrincipalArnToRoleArn(iamPrincipalArn);
+        // if the arn is not already in 'role' format, and cannot be found,
+        // then try checking for the generic "arn:aws:iam::0000000000:role/foo" format
+        if (!iamRole.isPresent() && !awsIamRoleArnParser.isRoleArn(iamPrincipalArn) ) {
+            logger.debug("Could not find IAM role in principal format, trying 'role' format...");
+            final String iamPrincipalInRoleFormat = awsIamRoleArnParser.convertPrincipalArnToRoleArn(iamPrincipalArn);
 
-                iamRole = awsIamRoleDao.getIamRole(iamPrincipalInRoleFormat);
-            }
-        } catch(Exception e) {
-            logger.debug("Failed to convert principal ARN to role ARN");
+            iamRole = awsIamRoleDao.getIamRole(iamPrincipalInRoleFormat);
         }
 
         return iamRole;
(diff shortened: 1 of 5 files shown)


src/main/java/com/nike/cerberus/service/AuthenticationService.java
```diff
@@ -341,6 +341,25 @@ private VaultAuthResponse generateToken(final String username, final Set<String>
...
+        try {
+            if (! awsIamRoleArnParser.isRoleArn(iamPrincipalArn)) {
+                logger.debug("Could not find IAM role in principal format, trying 'role' format...");
+                final String iamPrincipalInRoleFormat = awsIamRoleArnParser.convertPrincipalArnToRoleArn(iamPrincipalArn);
+
+                final Set<String> additionalPolicies = buildPolicySet(iamPrincipalInRoleFormat);
+                allPolicies.addAll(additionalPolicies);
+            }
+        } catch (Exception e) {
+            logger.debug("Failed to convert principal ARN to role ARN");
```
tlisonbee: Exception should go in log message

src/main/java/com/nike/cerberus/service/AuthenticationService.java
```diff
@@ -368,29 +387,29 @@ private VaultAuthResponse generateToken(final String username, final Set<String>
...
             throw ApiException.newBuilder()
                     .withApiErrors(DefaultApiError.AUTH_IAM_PRINCIPAL_INVALID)
-                    .withExceptionMessage(String.format("The role: %s was not configured for any SDB",
-                            credentials.getIamPrincipalArn()))
+                    .withExceptionMessage(String.format("The role: %s was not configured for any SDB", iamPrincipalArn))
                     .build();
         }
 
-        final Optional<AwsIamRoleKmsKeyRecord> kmsKey = awsIamRoleDao.getKmsKey(iamRole.get().getId(), credentials.getRegion());
+        final Optional<AwsIamRoleKmsKeyRecord> kmsKey = awsIamRoleDao.getKmsKey(iamRole.get().getId(), region);
```
tlisonbee: nitpick: creating an extra variable for 'region' doesn't improve readability

src/main/java/com/nike/cerberus/service/AuthenticationService.java
```diff
@@ -467,4 +485,23 @@ protected String getKeyId(IamPrincipalCredentials credentials) {
...
         return metadata;
     }
+
+    protected Optional<AwsIamRoleRecord> findIamRoleAssociatedWithSdb(final String iamPrincipalArn) {
+        Optional<AwsIamRoleRecord> iamRole = awsIamRoleDao.getIamRole(iamPrincipalArn);
+
+        try {
+            // if the arn is not already in 'role' format, and cannot be found,
+            // then try checking for the generic "arn:aws:iam::0000000000:role/foo" format
+            if ( !( awsIamRoleArnParser.isRoleArn(iamPrincipalArn) || iamRole.isPresent() ) ) {
```
tlisonbee: Boolean logic should be simplified

src/main/java/com/nike/cerberus/service/AuthenticationService.java
```diff
@@ -467,4 +485,23 @@ protected String getKeyId(IamPrincipalCredentials credentials) {
...
+                logger.debug("Could not find IAM role in principal format, trying 'role' format...");
+                final String iamPrincipalInRoleFormat = awsIamRoleArnParser.convertPrincipalArnToRoleArn(iamPrincipalArn);
+
+                iamRole = awsIamRoleDao.getIamRole(iamPrincipalInRoleFormat);
+            }
+        } catch(Exception e) {
+            logger.debug("Failed to convert principal ARN to role ARN");
```
tlisonbee: Log exception

src/main/java/com/nike/cerberus/service/AuthenticationService.java
```diff
@@ -467,4 +485,23 @@ protected String getKeyId(IamPrincipalCredentials credentials) {
 
         return metadata;
     }
+
+    protected Optional<AwsIamRoleRecord> findIamRoleAssociatedWithSdb(final String iamPrincipalArn) {
+        Optional<AwsIamRoleRecord> iamRole = awsIamRoleDao.getIamRole(iamPrincipalArn);
+
+        try {
```
tlisonbee: Not sure we should have this try/catch or the one you added above.  What is the purpose?

As far as I can tell, if there is a bug in the code it will let things still partially function, but it would be better for the bug to be obvious so we can fix it, instead of obscure and hard to track down.

Though, I might have misunderstood the intention here.  Thanks.

src/main/java/com/nike/cerberus/util/AwsIamRoleArnParser.java
```diff
@@ -19,47 +19,76 @@
...
  */
-// TODO: remove
 public class AwsIamRoleArnParser {
 
     public static final String AWS_IAM_ROLE_ARN_TEMPLATE = "arn:aws:iam::%s:role/%s";
 
     public static final String AWS_IAM_ROLE_ARN_REGEX = "^arn:aws:iam::(?<accountId>\\d+?):role/(?<roleName>.+)$";
 
-    public static final String AWS_IAM_PRINCIPAL_ARN_REGEX = "^arn:aws:(iam|sts)::.+$";
+    public static final String AWS_IAM_PRINCIPAL_ARN_REGEX = "^arn:aws:(iam|sts)::(?<accountId>\\d+?):(?!group).+?/(?<roleName>.+)$";
```
tlisonbee: This might be a nitpick but it seems like all of the regex in this class should probably be private.

src/main/java/com/nike/cerberus/util/AwsIamRoleArnParser.java
```diff
@@ -19,47 +19,76 @@
...
+
+    public String getRoleName(final String roleArn) {
+
+        return getNamedGroupFromPattern(IAM_ROLE_ARN_PATTERN, "roleName", roleArn);
 
-        return iamRoleArnMatcher.group("accountId");
     }
 
-    public String getRoleName(String roleArn) {
+    public boolean isRoleArn(final String arn) {
```
tlisonbee: Javadoc with an example is nice.

src/main/java/com/nike/cerberus/util/AwsIamRoleArnParser.java
```diff
@@ -19,47 +19,76 @@
...
+
+        final Matcher iamRoleArnMatcher = IAM_ROLE_ARN_PATTERN.matcher(arn);
+
+        return iamRoleArnMatcher.find();
+    }
+
+    public String convertPrincipalArnToRoleArn(final String principalArn) {
```
tlisonbee: Javadoc with an example is nice.

src/main/java/com/nike/cerberus/util/AwsIamRoleArnParser.java
```diff
@@ -19,47 +19,76 @@
...
+
+        if (isRoleArn(principalArn)) {
+            return principalArn;
+        }
+
+        final boolean isAssumedRole = GENERIC_ASSUMED_ROLE_PATTERN.matcher(principalArn).find();
+        final Pattern patternToMatch = isAssumedRole ? IAM_ASSUMED_ROLE_ARN_PATTERN : IAM_PRINCIPAL_ARN_PATTERN;
+
+        final String accountId = getNamedGroupFromPattern(patternToMatch, "accountId", principalArn);
+        final String roleName = getNamedGroupFromPattern(patternToMatch, "roleName", principalArn);
```
tlisonbee: I like how you refactored this into a method but it seems slightly off that we're calling find() and match() twice with the same pattern and input.  I'd be tempted to unroll the method.

src/main/java/com/nike/cerberus/service/AuthenticationService.java
sdford: Done

src/main/java/com/nike/cerberus/service/AuthenticationService.java
sdford: Done

src/main/java/com/nike/cerberus/service/AuthenticationService.java
sdford: Done

src/main/java/com/nike/cerberus/util/AwsIamRoleArnParser.java
sdford: Done

src/main/java/com/nike/cerberus/util/AwsIamRoleArnParser.java
sdford: Done

src/main/java/com/nike/cerberus/util/AwsIamRoleArnParser.java
sdford: Done

src/main/java/com/nike/cerberus/service/AuthenticationService.java
sdford: @tlisonbee The try/catch is a nice to have that allows any principal to authenticate, so long as it has assumed an IAM role with permissions.

I suspect that this code will more likely error due to user input, rather than a bug in this code (as it is unit tested). However, if the try portion fails because of a bug we are still logging the error and will know exactly why the code failed.

sdford: After talking to you about it, I think you're right.

Done.

src/main/java/com/nike/cerberus/util/AwsIamRoleArnParser.java
sdford: After talking to you, I'm going to leave this.

src/main/java/com/nike/cerberus/service/AuthenticationService.java
```diff
@@ -402,6 +402,27 @@ private VaultAuthResponse generateToken(final String username, final Set<String>
     }
 
     /**
+     * Build a set of policies that is allowed for the specific IAM principal (e.g. an assumed user, or instance profile)
```
fieldju: This comment doesn't explain this very well.

    Builds the complete set of policies for an IAM principal by ensuring assumed roles and instance profiles are also bound to there respective roles. 

    ex: An instance profile arn arn:aws:iam::1111111111:instance-profile/example or an assumed role arn:aws:sts::1111111111:assumed-role/example/i-abcdef will get policies that were set up specifically for the authenticating arn and it will also get the policies for the base role arn:aws:iam::1111111111:role/example.

    This is done to copy the way IAM policies work in the AWS ecosystem.

src/main/java/com/nike/cerberus/service/AuthenticationService.java
```diff
@@ -402,6 +402,27 @@ private VaultAuthResponse generateToken(final String username, final Set<String>
...
+     * as well as the generic role ARN (i.e. arn:aws:iam::1111111111:role/example)
+     * @param iamPrincipalArn - The given IAM principal ARN during authentication
+     * @return - List of all policies the given ARN has access to
+     */
+    protected Set<String> buildCompleteSetOfPolicies(final String iamPrincipalArn) {
+
+        final Set<String> allPolicies = buildPolicySet(iamPrincipalArn);
+
+        if (! awsIamRoleArnParser.isRoleArn(iamPrincipalArn)) {
+            logger.debug("Could not find IAM role in principal format, trying 'role' format...");
```
fieldju: This log statement is confusing, because I can authenticate with an instance-profile arn and have 2 SDBs set up in Cerberus one the references only the instance-profile arn and another that references the base role.

Might be better to be something like

    Detected non-role ARN, attempting to collect policies for principals base role

IAM Role ARNs, Instance Profile ARNs, Assumed Role ARNs are all principals so the language around an IAM Role vs a "principal format" is confusing.

src/main/java/com/nike/cerberus/util/AwsIamRoleArnParser.java
```diff
@@ -19,47 +19,98 @@
...
+     */
+    public boolean isRoleArn(final String arn) {
+
+        final Matcher iamRoleArnMatcher = IAM_ROLE_ARN_PATTERN.matcher(arn);
+
+        return iamRoleArnMatcher.find();
+    }
+
+    /**
+     * Converts a principal ARN (e.g. 'arn:aws:iam::0000000000:instance-profile/example') to a 'role' ARN,
```
fieldju: Again the language around this is confusing to me because in this situation isn't a role and an instance-profile both principals?

fieldju:
I am still digging through the PR, but

> A principal in computer security is an entity that can be authenticated by a computer system or network. It is referred to as a security principal in Java and Microsoft literature.

I have issues with the way you are using 'principal' to refer to some authenticatable identities and not others.

src/main/java/com/nike/cerberus/service/AuthenticationService.java
sdford: @fieldju ok, does adding the last line you put clear it up? or should I rephrase the rest as well

src/main/java/com/nike/cerberus/service/AuthenticationService.java
sdford: I agree with the language you recommend above. I will change this

src/main/java/com/nike/cerberus/util/AwsIamRoleArnParser.java
sdford: I see from your other comments how this doc can be confusing. I will find a way to rephrase.

However, to answer your question: no. A role in AWS is a container for IAM policies, while a principal is an actor who "assumes" that role and all of its permissions

PR #42, opened 2017-05-16
Allow any property to be overridden by the props file generated by the CLI

fieldju:


fieldju:
http://google.github.io/guice/api-docs/latest/javadoc/com/google/inject/util/Modules.html#override-java.lang.Iterable-

fieldju:
> Returns a builder that creates a module that overlays override modules over the given modules. If a key is bound in both sets of modules, only the binding from the override modules is kept. If a single PrivateModule is supplied or all elements are from a single PrivateBinder, then this will overwrite the private bindings. Otherwise, private bindings will not be overwritten unless they are exposed. This can be used to replace the bindings of a production module with test bindings:

PR #46, opened 2017-06-12
Fix Authentication in OktaMFAAuthConnector

sdford:
Fixes the issue where authentication fails with the OktaMFAAuthConnector if MFA is not required in Okta for the user.

**Note: This fix depends on dashboard change: Nike-Inc/cerberus-management-dashboard#23

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -14,6 +14,6 @@
 # limitations under the License.
 #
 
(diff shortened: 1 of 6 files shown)


sdford:
Closing this in favor of a cleaner solution involving v3/auth.

PR #47, opened 2017-06-15
updated reverse proxy port from 9000 to 9001

melanahammel:


james-michael:
Readme should be updated as well

PR #52, opened 2017-07-05
Only allow user principals to refresh their token and limit the number…

fieldju:
… of refreshes

### DIFF
--- a/src/main/java/com/nike/cerberus/security/VaultAuthPrincipal.java
+++ b/src/main/java/com/nike/cerberus/security/VaultAuthPrincipal.java
@@ -51,6 +51,10 @@ public class VaultAuthPrincipal implements Principal {
 
     public static final String METADATA_KEY_AWS_REGION = "aws_region";
 
+    public static final String METADATA_KEY_TOKEN_REFRESH_COUNT = "refresh_count";
+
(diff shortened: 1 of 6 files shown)


fieldju:
- [ ] Unit test that IAM principals get a 403

- [ ] Unit test the refresh limit logic

src/main/java/com/nike/cerberus/security/VaultAuthPrincipal.java
```diff
@@ -61,12 +65,22 @@
...
         this.username = extractUsername(clientToken);
         this.isIamPrincipal = extractIsIamPrincipal(clientToken);
+        this.tokenRefreshCount = extractTokenRefreshCount(clientToken);
+    }
+
+    private Integer extractTokenRefreshCount(VaultClientTokenResponse clientToken) {
+        final Map<String, String> meta = clientToken.getMeta();
+        // if a Token that is the root token or created outside of CMS,
+        // then meta might be null and there will be no value set
+        return meta == null ?  0 : Integer.parseInt(meta.getOrDefault(METADATA_KEY_TOKEN_REFRESH_COUNT, "0"));
```
sdford: Can the token refresh count be manipulated client-side by the user?

fieldju:
No, metadata along with everything else in and auth token is immutable and
stored in Vault.

PR #55, opened 2017-08-01
Adding a limited IAM auth cache to reduce calls to KMS and Vault

tlisonbee:


src/main/resources/cms.conf
```diff
@@ -55,6 +55,10 @@ service.version="@@RELEASE@@"
 # This sets the maximum number of times a token can be refreshed
 auth.token.maxRefreshCount=24
 
+# Cache time for auth tokens in seconds.  This value limits the calls to AWS KMS service.
+# This value should be a small fraction of cms.iam.token.ttl.override (default is 1h)
+cms.iam.token.cache.ttl=60
```
fieldju: Would it make any sense to have the unit in the key, like in_seconds or something?

tlisonbee:
We've decided not to implement this.  

1. This didn't fix the current bottleneck in the system (though it did help us track it down)
2. There are problems with the implementation, it needs more thought.  Maybe we'll revisit later.

Closing pull request.

PR #57, opened 2017-08-09
Updating instructions for running locally using gradle tasks

tlisonbee:


tlisonbee:
Yes, made change suggested by @fieldju

PR #58, opened 2017-08-09
Bug fix: RequestSecurityValidator for Riposte was misconfigured

tlisonbee:


fieldju:
What are the consequences of setting that to false?

tlisonbee:
The explanation of this change is in the Javadoc (auto-copied from Riposte).  Basically, you return true or false so that Riposte knows where to insert this.

- True means your implementation is fast so it should be handled by the Netty worker thread
- False means your implementation is slow so it should be done on an async thread.
- If you aren't sure, you just have to performance test to determine the right path.

@nicmunroe helped me find this when I reached out to him.  It would definitely cause a performance problem with Riposte but unfortunately it is not our current bottleneck.  We're hitting something else before this.

PR #63, opened 2017-08-29
Adding Hysterix library for getting metrics on calls to Vault and KMS

tlisonbee:


tlisonbee:
This is a work in progress.  I'm going to leave this on a branch for now.

src/main/java/com/nike/cerberus/util/ArchaiusUtils.java
```diff
@@ -0,0 +1,34 @@
+package com.nike.cerberus.util;
+
+import com.netflix.config.ConfigurationManager;
+import com.netflix.config.DynamicPropertyFactory;
+import com.typesafe.config.Config;
+
+import java.util.Properties;
+
+public class ArchaiusUtils {
```
fieldju: IMO, it's a bit confusing to add yet another way of dealing with properties to this project. It would be nice if we used Archaius everywhere and had a nice polling provider, so we don't have to do a rolling reboot for config changes. 

I would propose we move away from Guice handling property binds and use Archaius exclusively.

tlisonbee: I looked at implementing HystrixPropertiesStrategy and I think this is actually cleaner and simpler for now.  Also, this is a step also towards using Archaius as the main source for configuration.  I agree, polling would be nice.

PR #65, opened 2017-08-31
Allow Metrics to be sent to SignalFx

sdford:
This change enables users to enable metrics to be sent to SignalFx. This is optional and only works when the metrics reporting for SignalFx is turned on.

Required properties (even when not enabled):
metrics.signalfx.reporting.enabled=
metrics.reportJvmMetrics=
cms.app.name=
service.version=

Required properties (when enabled):
metrics.signalfx.api_key=
metrics.signalfx.dimension.app_env=


src/main/java/com/nike/cerberus/config/MetricsConfigurationHelper.java
```diff
@@ -0,0 +1,210 @@
...
+     * This 'holder' class allows optional injection of SignalFx-specific properties that are only necessary when
+     * SignalFx metrics reporting is enabled.
+     *
+     * The 'optional=true' parameter to Guice @Inject cannot be used in combination with the @Provides annotation
+     * or with constructor injection.
+     *
+     * https://github.com/google/guice/wiki/FrequentlyAskedQuestions
+     */
+    static class SignalFxOptionalPropertyHolder {
+        @Inject(optional=true)
```
nicmunroe: I'm pretty sure you can inject optional stuff using `@Nullable`, even into constructors or `@Provides` methods. Try the standard `javax.inject.Inject` and `javax.inject.Named` annotations rather than the guice-specific versions, and also mark it with `@Nullable`. If that doesn't work let me know and I'll mess around with it - I'm confident I've done this before.

nicmunroe: Also note that there are sometimes several `@Nullable` annotations available, and that whichever one you use must have `@Retention(RetentionPolicy.RUNTIME)` in order for Guice to be able to see it. When in doubt, use `javax.annotation.Nullable`.

nicmunroe: Further info here: https://github.com/google/guice/wiki/UseNullable

nicmunroe: Hmm, more reading makes it look like it still might not work. Worth a shot.

sdford: @nicmunroe I thought so too, hoping that Guice would intelligently inject `null` if a property key does not exist in the `Named` map, but my testing showed that Guice still errors with `No implementation bound` even when @Nullable is applied.

Unfortunately `@Nullable` does not substitute for having some sort of provider, binding, or `Named` value declaration that tells Guice how to inject that variable.

PR #66, opened 2017-09-01
Enabling TLS by default

tlisonbee:


fieldju:
Will this require changes to the the Gateway to enable SSL?

sdford:
@fieldju From what I understand, this won't require updating the gateway, but it will require updating the CMS ELB listener.

Correct me if I'm wrong @tlisonbee

tlisonbee:
Yes, it will require a change to the CMS ELB and probably a manual step or two if you want zero downtime during the upgrade.

tlisonbee:
Merging into highlander branch instead of master

PR #68, opened 2017-09-25
Serve Dashboard through Riposte

sdford:
Serve the Dashboard using Riposte instead of S3 hosting

** Note: This change won't be official until the network routing rules are updated in the Gateway/NGINX configuration.


src/main/java/com/nike/cerberus/AssetResourceFileFactory.java
```diff
@@ -0,0 +1,80 @@
...
+
+    /**
+     * @return The contents of the given file in bytes
+     */
+    private static byte[] getFileContents(String filePath) {
+        try {
+            ClassLoader loader = AssetResourceFileFactory.class.getClassLoader();
+            return IOUtils.toByteArray(loader.getResourceAsStream(filePath));
+        } catch (NullPointerException | IOException ioe) {
+            logger.info("Could not read contents of file: " + filePath, ioe);
```
tlisonbee: Is logger.error better here?  And are you sure we want to swallow exceptions here?

src/main/java/com/nike/cerberus/domain/AssetResourceFile.java
```diff
@@ -0,0 +1,44 @@
+package com.nike.cerberus.domain;
```
tlisonbee: Needs license header

src/main/java/com/nike/cerberus/AssetResourceFileFactory.java
sdford: @tlisonbee No, that was a mistake. Thanks

src/main/java/com/nike/cerberus/domain/AssetResourceFile.java
sdford: Done

fieldju:
The only thing I don't like about this is there is no option to run the dashboard outside of the JVM which disables things like live reload and live updates, if you make one tweak in the dashboard you have to recompile java. We need a better way when we do dashboard development.

sdford:
@fieldju Good point. I updated `server.js` to allow you to either load the dashboard from the JAR or using Webpack using the `loadDashboardFromCms` variable near the top.

PR #74, opened 2017-10-09
refactor hystrix metrics logger to optionaly debug log metrics but al…

fieldju:
…so send metrics to codahale and use scheduled job framework


src/main/java/com/nike/cerberus/jobs/HystrixMetricsProcessingJob.java
```diff
@@ -0,0 +1,160 @@
...
+ */
+@Singleton
+public class HystrixMetricsProcessingJob extends Job {
+
+    private static final Logger log = LoggerFactory.getLogger(HystrixMetricsProcessingJob.class);
+
+    private final SignalFxAwareCodahaleMetricsCollector metricsCollector;
+
+    @Inject
+    public HystrixMetricsProcessingJob(SignalFxAwareCodahaleMetricsCollector metricsCollector) {
```
sdford: You want to use the generic `CodahaleMetricsCollector` class here in case SignalFx metrics are disabled.

src/main/resources/cms.conf
```diff
@@ -121,6 +117,12 @@ cms.jobs.configuredJobs=[
         "repeatCount":  -1,
         "repeatInterval": 10,
         "repeatTimeUnit": "minutes"
+    },
+    {
+        "jobClassName": "HystrixMetricsProcessingJob",
+        "repeatCount":  -1,
+        "repeatInterval": 1,
```
fieldju: @tlisonbee what should the default interval be 10 seconds?

src/main/java/com/nike/cerberus/jobs/HystrixMetricsProcessingJob.java
```diff
@@ -0,0 +1,161 @@
...
+                    metrics.getCommandKey().name(),
+                    isCircuitOpen,
+                    metrics.getExecutionTimeMean(),
+                    metrics.getExecutionTimePercentile(95.0),
+                    metrics.getExecutionTimePercentile(99.0),
+                    metrics.getExecutionTimePercentile(99.5),
+                    metrics.getHealthCounts()
+            );
+
+            String baseMetricName = "cerberus.hystrix.command" +
```
tlisonbee: This base name might be too long.  I forget the details but I believe the names are supposed to be kept shorter than you'd like for performance reasons and other system limitations.

src/main/resources/cms.conf
tlisonbee: I'm not sure, 10 at minimum

src/main/java/com/nike/cerberus/jobs/HystrixMetricsProcessingJob.java
```diff
@@ -0,0 +1,161 @@
...
+                    gauge.setValue(metrics.getExecutionTimePercentile(99.5)));
+
+            getOrCreateSettableGauge(baseMetricName + ".totalCount").ifPresent((gauge) ->
+                    gauge.setValue(metrics.getHealthCounts().getTotalRequests()));
+
+            getOrCreateSettableGauge(baseMetricName + ".errorCount").ifPresent((gauge) ->
+                    gauge.setValue(metrics.getHealthCounts().getErrorCount()));
+        }
+    }
+
```
tlisonbee: Did we check with anyone on an existing library or sample code we can use?

src/main/java/com/nike/cerberus/jobs/HystrixMetricsProcessingJob.java
```diff
@@ -0,0 +1,161 @@
...
+    public void processHystrixThreadPoolMetrics() {
+        for (HystrixThreadPoolMetrics metrics : HystrixThreadPoolMetrics.getInstances()) {
+            log.debug("threadPool:{}, rollingCounts[rejected:{}, executed:{}, maxActiveThreads:{}], cumulativeCounts[rejected:{}, executed:{}], {}",
```
tlisonbee: To test this I would keep the debug logging turned on so that you can see if the numbers are matching up with Sfx

src/main/java/com/nike/cerberus/jobs/HystrixMetricsProcessingJob.java
sdford: We asked @nicmunroe some initial questions. This is an adaptation from his response.

src/main/java/com/nike/cerberus/jobs/HystrixMetricsProcessingJob.java
fieldju: ok, I will turn on debug logging for this class in the log back config
- [ ] update logback

src/main/resources/cms.conf
fieldju: - [ ]  Start with 10 seconds

PR #83, opened 2017-11-06
Reject instance-profile ARNs because they do not work with KMS policies

tlisonbee:


tlisonbee:
Closing, will merge this change from master instead

PR #84, opened 2017-11-07
Bug fixes: key policy validation and ARN validation

tlisonbee:
1. key policy validation was checking for ARNs that aren't allowed in keys.
2. ARN validation in previous commit broke instance-profile ARNs that could be converted to role ARNs

### DIFF
--- a/src/main/java/com/nike/cerberus/domain/IamPrincipalCredentials.java
+++ b/src/main/java/com/nike/cerberus/domain/IamPrincipalCredentials.java
@@ -21,14 +21,14 @@ import org.hibernate.validator.constraints.NotBlank;
 
 import javax.validation.constraints.Pattern;
 
-import static com.nike.cerberus.util.AwsIamRoleArnParser.AWS_IAM_PRINCIPAL_ARN_REGEX;
+import static com.nike.cerberus.util.AwsIamRoleArnParser.AWS_IAM_PRINCIPAL_ARN_REGEX_ALLOWED;
 
 /**
  * Represents the IAM principal credentials sent during authentication.
  */
 public class IamPrincipalCredentials {
 
-    @Pattern(regexp = AWS_IAM_PRINCIPAL_ARN_REGEX, message = "AUTH_IAM_PRINCIPAL_INVALID")
+    @Pattern(regexp = AWS_IAM_PRINCIPAL_ARN_REGEX_ALLOWED, message = "AUTH_IAM_PRINCIPAL_INVALID")
     private String iamPrincipalArn;
 
     @NotBlank(message = "AUTH_IAM_PRINCIPAL_AWS_REGION_BLANK")
--- a/src/main/java/com/nike/cerberus/domain/IamPrincipalPermission.java
+++ b/src/main/java/com/nike/cerberus/domain/IamPrincipalPermission.java
@@ -24,7 +24,7 @@ import javax.validation.constraints.Pattern;
 import javax.validation.groups.Default;
 import java.time.OffsetDateTime;
 
-import static com.nike.cerberus.util.AwsIamRoleArnParser.AWS_IAM_PRINCIPAL_ARN_REGEX;
+import static com.nike.cerberus.util.AwsIamRoleArnParser.AWS_IAM_PRINCIPAL_ARN_REGEX_ALLOWED;
 
 /**
  * Represents a permission granted to an IAM role with regards to a safe deposit box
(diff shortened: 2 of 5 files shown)


src/main/java/com/nike/cerberus/service/KmsPolicyService.java
```diff
@@ -112,7 +112,7 @@ protected boolean consumerPrincipalIsAnArnAndNotAnId(String policyJson) {
                             StringUtils.equals(statement.getId(), CERBERUS_CONSUMER_SID) &&
                                     statement.getPrincipals()
                                             .stream()
-                                            .anyMatch(principal -> awsIamRoleArnParser.isRoleArn(principal.getId())));
+                                            .anyMatch(principal -> awsIamRoleArnParser.isArnThatCanGoInKeyPolicy(principal.getId())));
```
tlisonbee: There was a bug here where if anything other than a role ARN was in a policy we considered it invalid.  Really we just want to catch IDs so we want to allow any ARN type that can go in a policy.

src/main/java/com/nike/cerberus/util/AwsIamRoleArnParser.java
```diff
@@ -30,15 +30,43 @@
 
     public static final String AWS_IAM_ROLE_ARN_TEMPLATE = "arn:aws:iam::%s:role/%s";
 
-    public static final String AWS_IAM_PRINCIPAL_ARN_REGEX = "^arn:aws:(iam|sts)::(?<accountId>\\d+?):(role|user|federated-user|assumed-role).*?/(?<roleName>.+)$";
+    /**
```
tlisonbee: I realized my previous commit earlier today introduced a bug.  Basically, this regex was doing two things:
1. validating ARNs that should be allowed in the DB.
2. the regex was being used for generating role ARNs from other ARN types.

Since we still want to be able to generate role ARNs from instance-profile ARNs (for backwards compatibility) but we don't want them in the DB anymore that means we need two regexes here.

src/main/java/com/nike/cerberus/util/AwsIamRoleArnParser.java
```diff
@@ -92,7 +126,7 @@ public String convertPrincipalArnToRoleArn(final String principalArn) {
         }
 
         final boolean isAssumedRole = GENERIC_ASSUMED_ROLE_PATTERN.matcher(principalArn).find();
-        final Pattern patternToMatch = isAssumedRole ? IAM_ASSUMED_ROLE_ARN_PATTERN : IAM_PRINCIPAL_ARN_PATTERN;
+        final Pattern patternToMatch = isAssumedRole ? IAM_ASSUMED_ROLE_ARN_PATTERN : IAM_PRINCIPAL_ARN_PATTERN_ROLE_GENERATION;
```
tlisonbee: This method needs to still generate role ARNs for a handful of instance-profile ARNs

src/main/java/com/nike/cerberus/domain/IamPrincipalCredentials.java
```diff
@@ -21,14 +21,14 @@
...
-import static com.nike.cerberus.util.AwsIamRoleArnParser.AWS_IAM_PRINCIPAL_ARN_REGEX;
+import static com.nike.cerberus.util.AwsIamRoleArnParser.AWS_IAM_PRINCIPAL_ARN_REGEX_ROLE_GENERATION;
 
 /**
  * Represents the IAM principal credentials sent during authentication.
  */
 public class IamPrincipalCredentials {
 
-    @Pattern(regexp = AWS_IAM_PRINCIPAL_ARN_REGEX, message = "AUTH_IAM_PRINCIPAL_INVALID")
+    @Pattern(regexp = AWS_IAM_PRINCIPAL_ARN_REGEX_ROLE_GENERATION, message = "AUTH_IAM_PRINCIPAL_INVALID")
```
tlisonbee: More permissive regex allows existing instance-profile ARN auth

src/main/java/com/nike/cerberus/domain/IamPrincipalPermission.java
```diff
@@ -36,7 +36,7 @@
     @NotBlank(message = "IAM_ROLE_ROLE_ID_INVALID", groups = {Default.class, Updatable.class})
     private String roleId;
 
-    @Pattern(regexp = AWS_IAM_PRINCIPAL_ARN_REGEX, message = "SDB_IAM_PRINCIPAL_PERMISSION_ARN_INVALID", groups = {Default.class, Updatable.class})
+    @Pattern(regexp = AWS_IAM_PRINCIPAL_ARN_REGEX_ALLOWED, message = "SDB_IAM_PRINCIPAL_PERMISSION_ARN_INVALID", groups = {Default.class, Updatable.class})
```
tlisonbee: Restrictive regex prevents additional instance-profile ARNs - we don't want them used any more because we couldn't use them in KMS key policies so the only worked when we could convert to a role ARN

PR #88, opened 2017-11-08
Prevent wrong CMS AMI from being deployed

tlisonbee:
* Adding intentional incompatible migration script to protect against accidentally deploying the wrong CMS AMI in the wrong environment.


src/main/resources/com/nike/cerberus/migration/V1.5.0.0__intentional_incompatible_change.sql
```diff
@@ -0,0 +1,15 @@
+#
+# This is a dummy migration script.  It is was created to introduce an intentional incompatible
```
sdford: Typo here

PR #93, opened 2017-11-17
Parameterize which certificate to use, flatten s3 folder structure

fieldju:
CLI changes coming soon

### DIFF
--- a/src/main/java/com/nike/cerberus/server/config/guice/CmsGuiceModule.java
+++ b/src/main/java/com/nike/cerberus/server/config/guice/CmsGuiceModule.java
@@ -20,6 +20,7 @@ package com.nike.cerberus.server.config.guice;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.inject.AbstractModule;
 import com.google.inject.Injector;
+import com.google.inject.Key;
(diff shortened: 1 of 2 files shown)


src/main/java/com/nike/cerberus/server/config/guice/CmsGuiceModule.java
```diff
@@ -345,8 +346,10 @@ public SslContext sslContext(@Named("cms.ssl.protocolsEnabled") String protocols
                     .build();
         } else {
             logger.info("initializing SslContext using certificate from S3");
-            InputStream certificate = IOUtils.toInputStream(cmsEnvPropertiesLoader.getCertificate(), Charset.defaultCharset());
-            InputStream privateKey = IOUtils.toInputStream(cmsEnvPropertiesLoader.getPrivateKey(), Charset.defaultCharset());
+            String certificateName = injector.getInstance(Key.get(String.class, Names.named("cms.ssl.certificateName")));
```
sdford: Why not just inject this as a parameter?

fieldju: because it wont be present for local envs

PR #95, opened 2017-11-29
Add KPI metrics processing job

sdford:
I haven't tested that metrics get to the metrics service, but want to start getting eyes on it.

### DIFF
--- a/src/main/java/com/nike/cerberus/jobs/KpiMetricsProcessingJob.java
+++ b/src/main/java/com/nike/cerberus/jobs/KpiMetricsProcessingJob.java
@@ -1,0 +1,127 @@
+/*
+ * Copyright (c) 2017 Nike, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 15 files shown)


src/main/java/com/nike/cerberus/jobs/KpiMetricsProcessingJob.java
```diff
@@ -0,0 +1,117 @@
...
+                gauge.setValue(uniqueOwnerGroups));
+
+        getOrCreateSettableGauge("number_of_sdbs").ifPresent((gauge) ->
+                gauge.setValue(numSDBs));
+
+        getOrCreateSettableGauge("number_of_data_nodes").ifPresent((gauge) ->
+                gauge.setValue(numDataNodes));
+    }
+
+    private Optional<SettableLongGauge> getOrCreateSettableGauge(String name) {
```
fieldju: is this duplicated code at this point?

May be time to have a metrics service with this method.

src/main/resources/com/nike/cerberus/mapper/UserGroupMapper.xml
```diff
@@ -68,6 +68,24 @@
         UGP.ROLE_ID = #{roleId}
   </select>
 
+  <select id="getUniqueSDBOwners" resultType="UserGroupRecord">
```
fieldju: Do you actually need all that data or just a count?

PR #97, opened 2017-12-06
Allow restore from non-highlander Cerberus environment

sdford:


src/main/java/com/nike/cerberus/service/SecureDataService.java
```diff
@@ -120,7 +143,9 @@ public int getTotalNumberOfDataNodes() {
      *
      * @param subPath The sub path to delete all secrets that have paths that start with
      */
+    @Transactional
     public void deleteAllSecretsThatStartWithGivenPartialPath(String subPath) {
+        log.warn("Recursively deleting secrets under path: {}", subPath);
```
fieldju: nit: it's not recursive anymore

PR #100, opened 2017-12-08
Dashboard: Only allow owners to which the user belongs

sdford:
Also, fix local dashboard development.

### DIFF
--- a/dashboard/server.js
+++ b/dashboard/server.js
@@ -6,9 +6,8 @@ var config = require("./webpack.config.js");
 var reverseProxyPort = 9001
 var nodeServerPort = 8000
 var cmsPort = 8080
-var vaultPort = 8200
 
-var loadDashboardFromCms = true
+var loadDashboardFromCms = false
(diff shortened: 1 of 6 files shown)


dashboard/server.js
```diff
@@ -6,9 +6,8 @@ var config = require("./webpack.config.js");
 var reverseProxyPort = 9001
 var nodeServerPort = 8000
 var cmsPort = 8080
-var vaultPort = 8200
 
-var loadDashboardFromCms = true
+var loadDashboardFromCms = false
```
tlisonbee: Why this change?  Seems unrelated to the pull request.

Is false the better default setting?  I'm guessing it is better for hot-reload or similar.

sdford: Yes, that's correct about hot-reload. If you're testing the dashboard, you want Webpack to serve the page so that you don't have to rebuild CMS every time you make a change

fieldju: Yeah, I think it is for hot reloading, else you have to rebuild the jar and restart the webapp every time you adjust a css margin.

PR #110, opened 2018-01-16
API.md updates

tlisonbee:
- Adding secrets API, healthcheck endpoint, trigger job endpoint.
- Switching to X-Cerberus-Token
- Adding some example error responses
- Various clean-up

### DIFF
--- a/API.md
+++ b/API.md
@@ -73,6 +73,19 @@ This endpoint will take a Users credentials, validate them with configured Auth 
               }
             }
 
++ Response 401 (application/json)
+
+    + Body
+
+            {
+                "error_id":"ccc1cc1c-e111-11e1-11ce-111e11a111f1",
+                "errors": [
+                    {
+                        "code":99106,
+                        "message":"Invalid credentials"
+                    }
+                ]
+            }
 
 ## User MFA Check [/v2/auth/mfa_check]
 
@@ -183,7 +196,7 @@ This endpoint takes IAM ARN information and generates a base64 encoded KMS encry
 + Response 200 (application/json)
     + The response will be a simple JSON payload with the encrypted data
             {
-                "auth_data": "long-encrypted-string"
+                "auth_data": "xxxxxxxxxxxxxxxxxxxxx-long-encrypted-string-xxxxxxxxxxxxxxxxxxxxxxxxxxxx"
             }
     + Once you have the encrypted string, you need to make a call to AWS Key Management Service (KMS) to decrypt the response. The decrypted response will contain the body below with the token needed to access Cerberus
     + Body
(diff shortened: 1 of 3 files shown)


API.md
```diff
@@ -622,27 +652,117 @@ This endpoint allows a user to delete a safe deposit box that they own
...
+
+    + Body
+
+        {
+          "request_id" : "bbdd111111a2aaa3a",
+          "lease_id" : "",
+          "renewable" : false,
+          "lease_duration" : 3600,
+          "data" : {
+            "keys" : [ "path1", "path2", "path3" ]
```
fieldju: if a key ends with `/` ex: `path4/` then it is a virtual path and doesn't contain secure data.

Not sure if or how that should be incorporated into this example

API.md
```diff
@@ -153,10 +165,14 @@ refresh_count and max_refresh_count can be used to determine when a re-authentic
 
 ### Authenticate with Cerberus as an App [POST]
 
-This endpoint takes IAM ARN information and generates an base 64 encoded KMS encrypted payload of the below. The ARN if registered with an SDB will have kms decrypt permissions on the KMS key that the payload was enrypted with.
+This endpoint takes IAM ARN information and generates a base64 encoded KMS encrypted payload.
```
fieldju: The response object from this request is as follows

    {
      "auth_data": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
    }

The value of "auth_data" is the kms encrypted base64 encoded payload that has a structure of 

    {
       "client_token" : "e5fd901b-bc65-71e7-a214-6066fef1e918",
       "policies" : [ "foo-bar-read", "lookup-self" ],
       "metadata" : {
         "aws_region" : "us-west-2",
         "iam_principal_arn" : "arn:aws:iam::111111111:role/fake-role"
         "username" : "arn:aws:iam::111111111:role/fake-role"
         "is_admin": "false",
         "groups": "registered-iam-principals"
       },
       "lease_duration" : 3600,
       "renewable" : true
     }

I feel like that could be documented better here and was always a point of confusion for people trying to make new clients or do things manually.

API.md
```diff
@@ -121,7 +132,8 @@ refresh_count and max_refresh_count can be used to determine when a re-authentic
 
     + Headers
 
-            X-Vault-Token: 7f6808f1-ede3-2177-aa9d-45f507391310
+            X-Cerberus-Token: 7f6808f1-ede3-2177-aa9d-45f507391310
```
fieldju: Nit: it doesn't really matter but tokens no longer contain non-alphanumeric chars, no more hyphens.

tlisonbee: Updated

API.md
tlisonbee: There were two examples that were different from each other.  I updated them to be the same and used the better example.

PR #117, opened 2018-01-26
Feature/audit logging

fieldju:
- Creates an optional audit log that is disabled by default that will log messages in a flattened JSON object that is Athena friendly.
- If the properties are set in environment.properties the logback configuration will automatically copy the audit logs that get rolled every 5 mintues to S3

example if the following are set

```
cms.event.processors.com.nike.cerberus.event.processor.AuditLogProcessor=true
cms.audit.bucket="athena-test-xxxxxx"
cms.audit.bucket_region="us-west-2"
```

Then the processor and s3 copying will be enabled.

un-minified example of audit log message
```json
{
  "event-timestamp":"2018-01-26 11:35:35",
  "principal-name":"arn:aws:iam::xxxxxxx:role/dev-cerberus-health-check-CerberusHealthCheckRole-xxxxxxx",
  "principal-type":"IAM",
  "principal-token-created":"2018-01-26 11:35:35",
  "principal-token-expires":"2018-01-27 12:35:35",
  "principal-is-admin":"false",
  "ip-address":"22.22.22.22",
  "x-forwarded-for":"33.33.33.33",
  "client-version":"Unknown",
  "http-method":"GET",
  "path":"/v1/secret/app/health-check-bucket/healthcheck",
  "action":"Reading secret with path: /v1/secret/app/health-check-bucket/healthcheck",
  "name":"ReadSecureData Endpoint Called",
  "sdb-name-slug":"health-check-bucket",
  "originating-class":"ReadSecureData"
}
```

The intent is to create a command in the CLI that will create the bucket and configure the Athena table and enable the processor.

### DIFF
--- a/src/main/java/com/nike/cerberus/endpoints/CustomizableAuditData.java
+++ b/src/main/java/com/nike/cerberus/endpoints/CustomizableAuditData.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright (c) 2018 Nike, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.nike.cerberus.endpoints;
+
+public class CustomizableAuditData {
+
+    private String description;
+    private String sdbNameSlug;
+
+    public String getDescription() {
+        return description;
+    }
+
+    public CustomizableAuditData setDescription(String description) {
+        this.description = description;
+        return this;
+    }
+
+    public String getSdbNameSlug() {
+        return sdbNameSlug;
+    }
+
+    public CustomizableAuditData setSdbNameSlug(String sdbNameSlug) {
+        this.sdbNameSlug = sdbNameSlug;
+        return this;
+    }
+}
--- a/src/main/java/com/nike/cerberus/event/processor/AuditLogProcessor.java
+++ b/src/main/java/com/nike/cerberus/event/processor/AuditLogProcessor.java
@@ -51,29 +51,29 @@ public class AuditLogProcessor implements EventProcessor {
             Optional<CerberusPrincipal> cerberusPrincipal = event.getPrincipalAsCerberusPrincipal();
 
             ImmutableMap<String, String> flattenedAuditEvent = ImmutableMap.<String, String>builder()
-                    .put("event-timestamp", event.getTimestamp().format(ATHENA_DATE_FORMATTER))
-                    .put("principal-name", event.getPrincipalName())
-                    .put("principal-type", cerberusPrincipal
+                    .put("event_timestamp", event.getTimestamp().format(ATHENA_DATE_FORMATTER))
+                    .put("principal_name", event.getPrincipalName())
+                    .put("principal_type", cerberusPrincipal
                             .map(p -> cerberusPrincipal.get().getPrincipalType().getName()).orElse(AuditableEvent.UNKNOWN))
-                    .put("principal-token-created", cerberusPrincipal
+                    .put("principal_token_created", cerberusPrincipal
                             .map(p -> cerberusPrincipal.get().getTokenCreated().format(ATHENA_DATE_FORMATTER))
                             .orElseGet(() -> OffsetDateTime.parse(PARTY_LIKE_ITS_99, ISO_OFFSET_DATE_TIME).format(ATHENA_DATE_FORMATTER)))
-                    .put("principal-token-expires", cerberusPrincipal
+                    .put("principal_token_expires", cerberusPrincipal
                             .map(p -> cerberusPrincipal.get().getTokenExpires().format(ATHENA_DATE_FORMATTER))
                             .orElseGet(() -> OffsetDateTime.parse(PARTY_LIKE_ITS_99, ISO_OFFSET_DATE_TIME).format(ATHENA_DATE_FORMATTER)))
-                    .put("principal-is-admin", cerberusPrincipal
+                    .put("principal_is_admin", cerberusPrincipal
                             .map(p -> String.valueOf(p.isAdmin())).orElseGet(() -> String.valueOf(false)))
-                    .put("ip-address", event.getIpAddress())
-                    .put("x-forwarded-for", event.getxForwardedFor())
-                    .put("client-version", event.getClientVersion())
-                    .put("http-method", event.getMethodAsString())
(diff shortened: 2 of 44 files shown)


src/main/java/ch/qos/logback/core/rolling/AuditLogsS3FixedWindowRollingPolicy.java
```diff
@@ -0,0 +1,120 @@
...
+        if (!rolledLogFile.exists() || rolledLogFile.length() == 0) {
+            return;
+        }
+
+        executor.execute(() -> {
+            try {
+                getAmazonS3().putObject(getBucket(), ENV_NAME + "/audit-logs/" + rolledLogFile.getName(), rolledLogFile);
+                rolledLogFile.delete();
+            } catch (Exception ex) {
+                ex.printStackTrace();
```
tlisonbee: shouldn't this be a logger?

tlisonbee: Also, do we want to catch here?

src/main/java/com/nike/cerberus/endpoints/CustomizableAuditData.java
```diff
@@ -0,0 +1,41 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.nike.cerberus.endpoints;
+
+public class CustomizableAuditData {
```
tlisonbee: javadoc

src/main/java/ch/qos/logback/core/rolling/AuditLogsS3FixedWindowRollingPolicy.java
fieldju: You don't have access to the logger in side the log back internals

src/main/java/com/nike/cerberus/event/processor/AuditLogProcessor.java
```diff
@@ -0,0 +1,94 @@
...
+                    .put("action", event.getAction())
+                    .put("name", event.getName())
+                    .put("sdb-name-slug", event.getSdbNameSlug())
+                    .put("originating-class", event.getOriginatingClass())
+                    .build();
+
+            try {
+                auditLogger.info(om.writeValueAsString(flattenedAuditEvent));
+            } catch (JsonProcessingException e) {
+                LoggerFactory.getLogger(org.slf4j.Logger.ROOT_LOGGER_NAME).error("failed to log audit event", e);
```
tlisonbee: Is that really the logger you want?

fieldju: Yes, want the errors going to the main log appender and not the audit log

src/main/resources/logback.groovy
```diff
@@ -16,15 +16,14 @@
...
-import ch.qos.logback.core.rolling.FixedWindowRollingPolicy
-import ch.qos.logback.core.rolling.RollingFileAppender
-import ch.qos.logback.core.rolling.SizeBasedTriggeringPolicy
+import ch.qos.logback.core.util.FileSize
+import com.nike.cerberus.util.ConfigUtils
 
 import static ch.qos.logback.classic.Level.*
 import static ch.qos.logback.core.spi.ContextAware.addInfo
 
+System.setProperty("logback.debug", "true")
```
tlisonbee: We don't want debug do we?

src/main/java/com/nike/cerberus/event/processor/AuditLogProcessor.java
fieldju: I mean "ROOT" is fine with me, but It can be any string other than this.getClass()

src/main/resources/logback.groovy
fieldju: I just enabled this so I can get more info in dev2 and bake an AMI

src/main/resources/logback.groovy
```diff
@@ -193,8 +192,7 @@ logger("com.nike.metrics.newrelic.NewRelicReporter", WARN, allAsyncAppendersArra
 
 logger("com.nike.cerberus", INFO, allAsyncAppendersArray, false)
 
-// TODO revert before merge to master
-logger("com.nike.cerberus.jobs.HystrixMetricsProcessingJob", DEBUG, allAsyncAppendersArray, false)
+logger("com.nike.cerberus.jobs.HystrixMetricsProcessingJob", INFO, allAsyncAppendersArray, false)
```
tlisonbee: What was "revert before merge to master"?

Are there other bad changes in here?

Is this highlander?  Things we need to update?

tlisonbee:
This seems worth performance testing before merging.

I definitely would not want to merge till after next Monday.

src/main/resources/logback.groovy
fieldju: That was referencing the LOG level of DEBUG, I switched it back to INFO

src/main/resources/logback.groovy
tlisonbee: Are you removing then?

src/main/java/ch/qos/logback/core/rolling/AuditLogsS3FixedWindowRollingPolicy.java
tlisonbee: If that is the case, how about also adding a System.err with error message?  Also, maybe a comment?

fieldju:
Yeah, I agree about performance testing it, it is all async so it shouldn't have an impact, but I will verify. 

I am also verifying that the log rotation to S3 is working as I expect.

tlisonbee:
Also, maybe something needs to be added to docs or README?

tlisonbee:
It seems like the kind of change that doesn't hurt to let sit in a test environment for awhile in case issues show up: memory leaks, edge cases, etc.

I went through the code quickly but I also wonder about error scenarios, I'm guessing S3 client has some retry built-in, but also what happens if S3 is down for 2 hours?

fieldju:
Yeah letting it sit for a bit sounds reasonable. I am still UAT'ing the log rolling pieces to make sure everything is working as I expect, once I get that taken care of I will remove that debug statement and do a perf test.

As for S3 being down, as it is right now and I am open to idea's the AuditLogsS3FixedWindowRollingPolicy will attempt to upload the rolled file and then delete the local copy, if some part of that fails it just moves on and never attempts to do anything else with that file.
so if there is a network blip you lose 5 minutes of logs.

logback is configured to keep 100 rolled logs around before it deletes things itself, so in an S3 outage there would be 100 * 5 = 8.3 hours or 10gb (which ever is first) worth of logs around before they start getting deleted locally, with manual intervention required to get them in the S3 bucket when it comes back online.

Lets talk more Monday in person.

Cheers

src/main/resources/logback.groovy
fieldju: After perf testing, you have to have this enabled if you want any sort of logging or std out/err from these classes.

PR #120, opened 2018-01-30
prioritize KMS in the current region first

fieldju:


src/main/java/com/nike/cerberus/service/EncryptionService.java
```diff
@@ -141,12 +144,21 @@ private void validateEncryptionContext(ParsedCiphertext parsedCiphertext, String
...
                 .collect(Collectors.toList());
         return (MasterKeyProvider<KmsMasterKey>) MultipleProviderFactory.buildMultiProvider(providers);
     }
 
+    protected static List<String> getSortedArnListByCurrentRegion(List<String> cmkArns, Region currentRegion) {
+        List<String> newList = new LinkedList<>();
+        cmkArns.stream().filter(s -> s.contains(currentRegion.getName())).forEach(newList::add);
+        cmkArns.removeAll(newList);
+        newList.addAll(cmkArns);
+        return newList;
```
tlisonbee: This can be written better.  It could just be a sort method with a comparison.

PR #124, opened 2018-02-06
Update IAM ARN pattern to disallow ending in whitespace, add migratio…

fieldju:
…n script to trim whitespace from existing data


james-michael:
+1

PR #125, opened 2018-02-06
Created Athena service to automatically add partitions to audit table

fieldju:


src/main/java/com/nike/cerberus/service/AthenaService.java
```diff
@@ -0,0 +1,78 @@
...
+                        bucket,
+                        year, month, day, hour);
+
+                AmazonAthena athena = athenaClientFactory.getClient(region);
+
+                StartQueryExecutionResult result = athena.startQueryExecution(new StartQueryExecutionRequest()
+                        .withQueryString(query)
+                        .withResultConfiguration(new ResultConfiguration().withOutputLocation(String.format("s3://%s/results/", bucket)))
+                );
+                log.info("Started query: '{}' to add partition: '{}' to table: '{}'", result.getQueryExecutionId(), partition, table);
```
sdford: Any reason this should be info and not debug?

It seems like this will get noisy quickily, and I don't think we care about this being in the log.

fieldju: I can make it debug, it's only getting logged every 5 minutes, but I had that for testing anyways, which I am done with.

PR #130, opened 2018-02-13
Split up the manage SDB page to have tabs and be ready to add new com…

fieldju:
…ponents

image

image

image

image

image


sdford:
Are you planning to revert the requirement to delete all secrets before deleting an SDB as a part of this change?

fieldju:
> Are you planning to revert the requirement to delete all secrets before deleting an SDB as a part of this change?

Yes, I think now that the settings are on a new tab and the buttons are labeled and the dialog is clear and makes you type in the name of the SDB, that deleting the secrets is no longer needed to stop people from accidentally deleting and SDB when they think they are deleting a path / node.

sdford:
Should it say `verifiedSdbName` in camel case in the delete SDB pop-up?

Also, that text seems a little long. What do you think of a <h2>Are you sure you want to delete this SDB<h2>, with the description text smaller underneath?

fieldju:
> Should it say verifiedSdbName in camel case in the delete SDB pop-up?
No, which is why there is already a commit in this PR fixing that.

PR #139, opened 2018-02-20
Feature/ux ui refreshments

fieldju:
image

image


sdford:
Does it still say "secret" in the placeholder for adding the path, or is that an old picture?

sdford:
Otherwise looks great!

fieldju:
> Does it still say "secret" in the placeholder for adding the path, or is that an old picture?

@sdford It does, let me change that real quick

PR #141, opened 2018-02-22
some of the existing clients expect writing secure data to accept PUT…

fieldju:
… as well as POST, when we changed arch we made write only listen to POST


tlisonbee:
Looks fine but what is this build error?

fieldju:
I broke some tests, let me update the unit tests for that Enum.

PR #145, opened 2018-04-09
Alter the name field in the user groups table so that it supports cas…

fieldju:
…e sensitivity


fieldju:
I can now add 

```
foo
Foo
fOo
```

as separate groups.

I also confirmed that being a member of foo does not grant access to an sdb that has granted permissions to Foo.

src/main/resources/com/nike/cerberus/migration/V1.6.2.0__make_user_groups_case_sensitive.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE USER_GROUP CHANGE NAME NAME VARCHAR(255) BINARY NOT NULL;
```
sdford: My understanding of this is that changing `NAME` to binary will force the database to use the values of the bytes in the string to compare.

Where I assume, before, all letter casing would be the same regardless of the byte letter values?

fieldju: Yes, this will now allow `foo` and `Foo` and `fOo` to be considered different values and not cause collisions in the index as we did with iam_roles.

PR #146, opened 2018-04-09
* add admin endpoint to override the owner of an SDB

mayitbeegh:


src/main/java/com/nike/cerberus/endpoints/admin/OverrideSdbOwner.java
```diff
@@ -0,0 +1,81 @@
...
+                                                           final SecurityContext securityContext) {
+
+        return CompletableFuture.supplyAsync(
+                AsyncNettyHelper.supplierWithTracingAndMdc(() -> overrideSdbOwner(request, securityContext), ctx),
+                longRunningTaskExecutor
+        );
+    }
+
+    private ResponseInfo<Void> overrideSdbOwner(RequestInfo<SDBMetadata> request, SecurityContext securityContext) {
+        CerberusPrincipal authPrincipal = (CerberusPrincipal) securityContext.getUserPrincipal();
```
fieldju: Add an audit event here, see any of the other endpoints for an example.

src/main/java/com/nike/cerberus/service/MetadataService.java
```diff
@@ -151,6 +152,19 @@ public String getSdbId(SDBMetadata sdbMetadata) {
         return id;
     }
 
+    /**
+     * Overrides the owner of the SDB. This is an admin function.
+     * @param name the name of the SDB
+     */
+    public void overrideOwner(String name, String newOwner, String adminUser) {
```
fieldju: Why is this in the metadata service?

Seems like a method that should in the safe deposit box service exclusively.

fieldju: name, seems like a bad variable name in this instance, when it is really the sdbName. We have multiple things in this code base that have name. immediately looking at the method name I thought maybe it was the owner name. Looking at the java doc, I can finally see what it is for.

Also the Javadoc is missing params.

fieldju: - [ ] self describing variable name
- [ ] fix java do

mayitbeegh: updated

mayitbeegh: updated

PR #147, opened 2018-04-11
Update DB for file support

sdford:
The first of a few PRs for file support.


src/main/java/com/nike/cerberus/domain/SecureDataType.java
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package com.nike.cerberus.domain;
+
+public enum SecureDataType {
```
fieldju: A quick description might be nice on this, its mostly self explanatory though.

fieldju:
looks good, my comment is that we refer to the plaintext payload as plaintext payload sometimes and unencrypted payload others.

We should choose to use plaintext / ciphertext or encrypted / decrypted it is a little confusing to use both.

My preference it to use plaintext and ciphertext.

src/main/resources/com/nike/cerberus/migration/V1.6.2.0__add_type_to_secure_data.sql
```diff
@@ -0,0 +1,22 @@
+ALTER TABLE SECURE_DATA
+  ADD COLUMN `TYPE` VARCHAR(255) NOT NULL,
+  ADD COLUMN SIZE_IN_BYTES INTEGER NOT NULL;
+
+UPDATE SECURE_DATA
+  SET `TYPE` = 'OBJECT',
+      SIZE_IN_BYTES = -1;
```
sdford: @fieldju I want to call this out directly and see if you have any objections/suggestions for default size value

sdford:
Sounds good. I changed the language to plaintext/ciphertext where it makes sense and added javadoc to the `SecureDataType` enum

PR #167, opened 2018-06-25
* fix issue where getPath() would return "sdb-name/null" instead of "…

mayitbeegh:
…sdb-name/"

### DIFF
--- a/src/main/java/com/nike/cerberus/domain/SecureDataRequestInfo.java
+++ b/src/main/java/com/nike/cerberus/domain/SecureDataRequestInfo.java
@@ -19,6 +19,8 @@ package com.nike.cerberus.domain;
 
 import com.nike.cerberus.security.CerberusPrincipal;
 
+import java.util.Optional;
+
 public class SecureDataRequestInfo {
     private String category;
     private String sdbSlug;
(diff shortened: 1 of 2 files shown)


src/main/java/com/nike/cerberus/domain/SecureDataRequestInfo.java
```diff
@@ -68,6 +70,6 @@ public SecureDataRequestInfo setPrincipal(CerberusPrincipal principal) {
     }
 
     public String getPath() {
-        return String.format("%s/%s", sdbSlug, subPath);
+        return String.format("%s/%s", sdbSlug, Optional.ofNullable(subPath).orElse(""));
```
sdford: Does this fix the issue?

It seems like if a user passes a path without a `/` at the end, then it will still not append the `/`

mayitbeegh: Before getPath(), there's this piece of code:
https://github.com/Nike-Inc/cerberus-management-service/blob/master/src/main/java/com/nike/cerberus/SecureDataRequestService.java#L110
If subPath == null then before this fix getPath() returns "sdb-name/null", after this fix getPath returns "sdb-name/"
If subPath == null and subPath doesn't have trailing slash, then getPath() returns the same "sdb-name/sub-path". And then a trailing slash is added in https://github.com/Nike-Inc/cerberus-management-service/blob/master/src/main/java/com/nike/cerberus/service/SecureDataService.java#L288

sdford: Got it. I misread the code

PR #172, opened 2018-07-24
Feature/orphan kms key cleanup job

fieldju:
Add orphan clean up job to delete orphaned kms keys that can get created from a race condition in the kms authentication flow, when an IAM role authenticates for the first time.


src/main/java/com/nike/cerberus/hystrix/HystrixKmsClient.java
```diff
@@ -2,20 +2,7 @@
...
-import com.amazonaws.services.kms.model.DescribeKeyResult;
-import com.amazonaws.services.kms.model.EncryptRequest;
-import com.amazonaws.services.kms.model.EncryptResult;
-import com.amazonaws.services.kms.model.GetKeyPolicyRequest;
-import com.amazonaws.services.kms.model.GetKeyPolicyResult;
-import com.amazonaws.services.kms.model.PutKeyPolicyRequest;
-import com.amazonaws.services.kms.model.PutKeyPolicyResult;
-import com.amazonaws.services.kms.model.ScheduleKeyDeletionRequest;
-import com.amazonaws.services.kms.model.ScheduleKeyDeletionResult;
+import com.amazonaws.services.kms.model.*;
```
sdford: star import

src/main/java/com/nike/cerberus/service/KmsService.java
```diff
@@ -463,4 +467,119 @@ protected boolean kmsPolicyNeedsValidation(AwsIamRoleKmsKeyRecord kmsKeyRecord,
...
+     * @param kmsCmkId The KMS CMK id.
+     * @param regionName The region.
+     * @return true if the kms key was created by CMS for the current env
+     */
+    private boolean wasKeyCreatedByKmsService(String kmsCmkId, String regionName) {
+        boolean wasCreatedByThis = false;
+
+        logger.debug("Downloading policy for key: {} in region: {}", kmsCmkId, regionName);
+
+        Optional<Policy> policyOptional = downloadPolicy(kmsCmkId, regionName, 0);
```
sdford: Are there any rate limit concerns in calling `GetKeyPolicy` n times for n keys every time this job is run?

sdford:
Did you say before you were adding integration tests for this?

I didn't notice them here.

fieldju:
```Did you say before you were adding integration tests for this?

I didn't notice them here.```

No, not for this feature, the integration tests I was talking about are here on this branch.
https://github.com/Nike-Inc/cerberus-management-service/commit/2f790738b137e208f2cb3b6c91852ac2321edd9d

src/main/java/com/nike/cerberus/service/KmsService.java
fieldju: Yes, which is why in the downloadPolicy method, I have retries with exp back off, it retires 12 times and backs off and eventually returns an empty optional if it still fails.

The side effect of missing a key, is that the orphaned key wouldn't get cleaned up till the next job run.

src/main/java/com/nike/cerberus/hystrix/HystrixKmsClient.java
fieldju: - [ ] Remove wild card import

src/main/java/com/nike/cerberus/service/KmsService.java
fieldju: When, I ran this during my user acceptance testing, it never got throttled or blocked from rate limiting.

get key policy has a limit of 30 per sec, and it has a little latency, so in practice with this being a single thread, I have not seen that limit being reached. How ever if this was run by more than one env at once it could hit that limit. but should be fine because of the retries with back off.

sdford: I was more wondering what happens to Cerberus while this job is running.

If the job is running and Cerberus needs to call `GetKeyPolicy` will Cerberus be rate limited? Will Cerberus also fail gracefully / retry like this method does?

sdford: Also, you bring up a good point about multiple environments running at the same time.

Maybe that's unlikely since the runtime of the job is based on deploy time of the EC2 instances?

PR #173, opened 2018-07-25
Allow users to grant SDB permissions to an entire AWS account

sdford:


tlisonbee:
I kind of imagined this functionality would be enabled with SQL change.  For example, looking up SDBs that a role has access to would just add an `OR` onto an existing SQL statement.

PR #174, opened 2018-08-08
* add v3 authentication endpoint that accepts a signed AWS STS get-ca…

mayitbeegh:
…ller-identity request and returns an unencrypted payload containing the access token


src/main/java/com/nike/cerberus/aws/sts/AwsStsHttpClient.java
```diff
@@ -0,0 +1,147 @@
...
+    public <M> M execute(final Map<String, String> headers,
+                            final Class<M> responseClass) {
+        try {
+            Request request = buildRequest(headers);
+            Response response =  httpClient.newCall(request).execute();
+            if (!response.isSuccessful()) {
+                final String msg = String.format("Failed to authenticate with AWS, error message: %s",
+                        response.body().string());
+
+                throw ApiException.newBuilder()
```
fieldju: You are throwing this error assuming what a 403/401?

There are many other 400-500 series things that can go wrong, always throwing AUTH_BAD_CREDENTIALS even if a rate limit is hit or a service is down seems wrong.

src/main/java/com/nike/cerberus/endpoints/authentication/AuthenticateStsIdentity.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    private ResponseInfo<AuthTokenResponse> authenticate(RequestInfo<Void> request) {
+        final String date = getDate(request);
+        final String headerXAmzDate = getHeaderXAmzDate(request);
+        final String headerXAmzSecurityToken = getHeaderXAmzSecurityToken(request);
+        final String headerAuthorization = getHeaderAuthorization(request);
+
+        if (date == null || headerAuthorization == null || headerXAmzDate == null || headerXAmzSecurityToken == null) {
+            throw new ApiException(DefaultApiError.GENERIC_BAD_REQUEST);
```
fieldju: This error message / error code is not useful. Shouldn't you create a new error?

src/main/java/com/nike/cerberus/service/AuthenticationService.java
```diff
@@ -224,6 +224,13 @@ public IamRoleAuthResponse authenticate(IamPrincipalCredentials credentials) {
         return authenticate(credentials, authPrincipalMetadata);
     }
 
+    public AuthTokenResponse keylessAuthenticate(final String iamPrincipalArn) {
```
fieldju: java docs on this new public method.

src/main/java/com/nike/cerberus/service/AuthenticationService.java
```diff
@@ -269,6 +276,15 @@ private IamRoleAuthResponse authenticate(IamPrincipalCredentials credentials, Ma
         return iamRoleAuthResponse;
     }
 
+    private AuthTokenResponse keylessAuthenticate(String iamPrincipalArn, Map<String, String> authPrincipalMetadata) {
```
fieldju: java docs on this new public method.

API.md
```diff
@@ -10,6 +10,53 @@ as well as in the [integration tests](https://github.com/Nike-Inc/cerberus-integ
 
 # Group Authentication
 
+## App Login v3 [/v3/auth/sts-identity]
```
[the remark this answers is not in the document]
mayitbeegh: It's v3 of auth, as it authenticate in a different manner and returns unencrypted token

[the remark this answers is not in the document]
fieldju: sts-identity is a brand new path, is it seems like it should be /v1/auth/sts-identity or but I can see you putting it in the /v2/ group for sane grouping.

/v3/ makes no sense to me, as this endpoint is completely independent from `/v1/auth/iam-principal`

src/main/java/com/nike/cerberus/aws/sts/AwsStsHttpClient.java
```diff
@@ -0,0 +1,155 @@
...
+     *
+     * @param headers     HTTP Headers to include in the request
+     * @param responseClass The class of the response object
+     * @return Response from the server
+     */
+    public <M> M execute(final Map<String, String> headers,
+                            final Class<M> responseClass) {
+        try {
+            Request request = buildRequest(headers);
+            Response response =  httpClient.newCall(request).execute();
```
tlisonbee: The AWS SDK would do retry with exponential backoff.  Why are we doing http client instead?

src/main/java/com/nike/cerberus/aws/sts/AwsStsHttpClient.java
```diff
@@ -0,0 +1,155 @@
...
+                        ));
+
+        return requestBuilder.build();
+    }
+
+    protected ApiException toApiException(IOException e) {
+        if (e instanceof SSLException
+                && e.getMessage() != null
+                && e.getMessage().contains("Unrecognized SSL message, plaintext connection?")) {
+            return ApiException.newBuilder()
```
tlisonbee: This is only possible with local dev?  Is that right?  Should we add a comment?

src/main/java/com/nike/cerberus/aws/sts/AwsStsHttpClient.java
tlisonbee: Why did we write our own STS client?

src/main/java/com/nike/cerberus/server/config/CmsConfig.java
```diff
@@ -109,6 +102,7 @@ protected CmsConfig(Config appConfig, PropertiesRegistrationGuiceModule properti
                 new BackstopperRiposteConfigGuiceModule(),
                 new CmsFlywayModule(),
                 new OneLoginGuiceModule(),
+                new AwsStsGuiceModule(),
```
tlisonbee: Why not put this in the CMS Guice Module?  

Most of these others are separated for a reason but this just seems like other code that is in the CMS Guice Module

src/main/java/com/nike/cerberus/service/AuthenticationService.java
```diff
@@ -224,6 +224,18 @@ public IamRoleAuthResponse authenticate(IamPrincipalCredentials credentials) {
         return authenticate(credentials, authPrincipalMetadata);
     }
 
+    /**
+     * Enables an IAM role to authenticate and get back an UNENCRYPTED payload
+     * @param iamPrincipalArn IAM role ARN
+     * @return Unencrypted auth response
+     */
+    public AuthTokenResponse keylessAuthenticate(final String iamPrincipalArn) {
```
tlisonbee: instead of keyless auth, how about stsAuth?  or some other phrase describing "what it is" instead of saying "what it is not".

API.md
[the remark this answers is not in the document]
tlisonbee: I think of it more as an additional auth style vs. version 3 of the existing auth.  The endpoint even has a new name.

src/main/java/com/nike/cerberus/server/config/CmsConfig.java
tlisonbee: I looked at the module after writing the above comment.  I don't feel that strongly either way.

src/main/java/com/nike/cerberus/aws/sts/AwsStsHttpClient.java
tlisonbee: Did you ever see this exception?

mayitbeegh: I saw this in the OneLoginHttpClient. I included it because it looks like potentially a common Exception which probably isn't the case with AWS but I kept it anyway. I can remove this.

src/main/java/com/nike/cerberus/aws/sts/AwsStsHttpClient.java
mayitbeegh: I can add retry. The error handling here is pretty vanilla, but at least for this get-caller-identity call the SDK isn't doing anything too special either.

src/main/java/com/nike/cerberus/service/AuthenticationService.java
mayitbeegh: good point

src/main/java/com/nike/cerberus/aws/sts/AwsStsHttpHeader.java
```diff
@@ -0,0 +1,91 @@
...
+ */
+
+package com.nike.cerberus.aws.sts;
+
+import com.google.common.base.Preconditions;
+import com.google.common.collect.Maps;
+
+import java.util.Map;
+
+import static com.nike.cerberus.aws.sts.AwsStsHttpHeaders.*;
```
sdford: star import

src/main/java/com/nike/cerberus/service/AuthenticationService.java
```diff
@@ -272,6 +284,15 @@ private IamRoleAuthResponse authenticate(IamPrincipalCredentials credentials, Ma
         return iamRoleAuthResponse;
     }
 
+    private AuthTokenResponse stsAuthenticate(String iamPrincipalArn, Map<String, String> authPrincipalMetadata) {
```
sdford: Does this need to be a separate method? Why not put this code in the other `stsAuthenticata`

src/main/java/com/nike/cerberus/service/AuthenticationService.java
```diff
@@ -626,6 +645,19 @@ protected AwsIamRoleKmsKeyRecord getKmsKeyRecordForIamPrincipal(final AwsIamRole
...
     }
 
+    /**
+     * Generate map of Token metadata that is common to all principals
+     *
+     * @param iamPrincipalArn - The authenticating IAM principal ARN
+     * @param region - The AWS region
+     * @return - Map of token metadata
+     */
+    protected Map<String, String> generateCommonIamPrincipalAuthMetadata(final String iamPrincipalArn, final String region) {
```
sdford: same. does this need to be a separate method? doesn't look it's used

PR #175, opened 2018-08-23
Fix okta verify mapping

melanahammel:
Fixed the mapping for device names. Now when users are asked to use MFA to log in, they will see an accurate dropdown list of each verification method they have. Push, call, and sms notifications are filtered from the list because we do not currently support them.

### DIFF
--- a/src/main/java/com/nike/cerberus/auth/connector/okta/OktaAuthConnector.java
+++ b/src/main/java/com/nike/cerberus/auth/connector/okta/OktaAuthConnector.java
@@ -44,6 +44,8 @@ public class OktaAuthConnector implements AuthConnector {
 
     private final OktaClientResponseUtils oktaClientResponseUtils;
 
+    private static final ImmutableSet UNSUPPORTED_OKTA_MFA_TYPES = ImmutableSet.of("push", "call", "sms");
+
     @Inject
     public OktaAuthConnector(final OktaApiClientHelper oktaApiClientHelper,
                              final OktaClientResponseUtils oktaClientResponseUtils) {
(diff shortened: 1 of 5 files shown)


src/main/java/com/nike/cerberus/auth/connector/okta/OktaAuthConnector.java
```diff
@@ -62,19 +64,28 @@ public AuthResponse authenticate(String username, String password) {
...
 
-            factors = oktaClientResponseUtils.getUserFactorsFromAuthResult(authResult);
+            final List<Factor> factors = oktaClientResponseUtils.getUserFactorsFromAuthResult(authResult)
+                    .stream()
+                    // Filter out Okta push, call, and sms because we don't currently support them.
+                    .filter(factor -> {
+                        String type = factor.getFactorType().toLowerCase();
+                        String provider = factor.getProvider();
+                        return ! (provider.equalsIgnoreCase("okta") &&
+                                ImmutableSet.of("push", "call", "sms").contains(type));
```
tlisonbee: This ImmutableSet is better as a `private static final` on the class with a descriptive name, e.g `UNSUPPORTED_OKTA_MFA_TYPES`

src/main/java/com/nike/cerberus/auth/connector/okta/OktaAuthConnector.java
```diff
@@ -62,19 +64,28 @@ public AuthResponse authenticate(String username, String password) {
...
                 StringUtils.equals(authResult.getStatus(), OktaClientResponseUtils.AUTHENTICATION_MFA_ENROLL_STATUS)) {
 
             authData.setStateToken(authResult.getStateToken());
             authResponse.setStatus(AuthStatus.MFA_REQUIRED);
 
-            factors = oktaClientResponseUtils.getUserFactorsFromAuthResult(authResult);
+            final List<Factor> factors = oktaClientResponseUtils.getUserFactorsFromAuthResult(authResult)
+                    .stream()
+                    // Filter out Okta push, call, and sms because we don't currently support them.
+                    .filter(factor -> {
```
tlisonbee: The code in this filter is slightly better if moved into another method `isSupportedFactor(factor)`.  It would read cleaner and is easier to write a unit test for.

PR #178, opened 2018-09-21
added sms and call backend functionality

melanahammel:
- updated endpoint to trigger an MFA challenge in addition to verifying MFA factor 
- added new AuthStatus "MFA_CHALLENGE"
- added new state handler case "handleMfaChallenge"


src/main/java/com/nike/cerberus/auth/connector/okta/statehandlers/InitialLoginStateHandler.java
```diff
@@ -22,25 +22,6 @@ public InitialLoginStateHandler(AuthenticationClient client, CompletableFuture<A
         super(client, authenticationResponseFuture);
     }
 
-    /**
```
fieldju: Why was this removed, what happens if an OKTA Org does not have MFA turned on, can a user still log in?

src/main/java/com/nike/cerberus/auth/connector/onelogin/OneLoginAuthConnector.java
```diff
@@ -63,6 +63,12 @@ public AuthResponse authenticate(String username, String password) {
         return authResponse;
     }
 
+    @Override
+    public AuthResponse triggerChallenge(String stateToken, String deviceId) {
+
+        throw new UnsupportedOperationException("Method call is not allowed.");
```
fieldju: This will be an uncaught error that will show up to a user as a generic internal server error, when this seems like a bad request to me.

Perhaps an API error with a usable error message here would be better.

src/main/java/com/nike/cerberus/service/AuthenticationService.java
```diff
@@ -195,7 +209,7 @@ public AuthResponse mfaCheck(final MfaCheckRequest mfaCheckRequest) {
         return authResponse;
     }
 
-    /**
+    /**F
```
fieldju: typo

src/main/java/com/nike/cerberus/auth/connector/onelogin/OneLoginAuthConnector.java
fieldju: - [ ] Make this an API error with good error message.

melanahammel: I changed the error from Unsupported Operation to an API error. I was unable to make it a 501 because the ProjectAPIErrorsTestBase that we use did not include that error, so I made it a 405 Method Not Allowed.

src/main/java/com/nike/cerberus/error/DefaultApiError.java
```diff
@@ -293,6 +293,11 @@
      */
     FACTOR_VALIDATE_FAILED(99243, "Failed to validate factor. Please try again or try a different factor.", SC_UNAUTHORIZED),
 
+    /**
+     * Call to trigger sms or call challenge for OneLogin is not implemented.
+     */
+    TRIGGER_CHALLENGE_NOT_IMPLEMENTED(99244, "Call to trigger sms or call challenge for OneLogin is not implemented.", SC_METHOD_NOT_ALLOWED),
```
fieldju: This should be 501 status code, please refer to https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html when choosing status codes.

fieldju: 405 Method Not Allowed
```The method specified in the Request-Line is not allowed for the resource identified by the Request-URI. The response MUST include an Allow header containing a list of valid methods for the requested resource.```

vs

501 Not Implemented
```The server does not support the functionality required to fulfill the request.```

Issue #183, opened 2019-01-24
Regex Parsing Bug on IAM Role Authentication


dylan-raithel-lookout:
For an IAM role that exists, and is attached to an SDB, Cerberus throws an authentication error. It looks like a parsing bug on the role name that leads to the name being suffixed with a space. When I remove and re-add the role from the SDB, save it, then edit it, I can confirm that it's adding another space.

Initial stack trace:

A Important error occured
com.amazonaws.services.kms.model.AWSKMSException: Unable to authenticate <role name with an extra trailing space> IAM role due to the following error(s):
Error ID:64081477-2add-4ccf-a042-d6627076474f
Code: 10
Message: An error occurred while fulfilling the request
Please refer to the Cerberus Java Client documentation.
If you are unsure of what this means please ask in the #cd channel on Slack. (Service: null; Status Code: 0; Error Code: null; Request ID: null)
	at com.lookout.cerberus.client.SecretStoreClient.authenticate(SecretStoreClient.java:110)
	at com.lookout.cerberus.client.SecretStoreClient.<init>(SecretStoreClient.java:50)
	at com.lookout.cerberus.client.SecretStoreClient.<init>(SecretStoreClient.java:58)
	at com.lookout.cerberus.client.Cerberus.initializeCerberusClient(Cerberus.java:82)
	at com.lookout.cerberus.client.Cerberus.fetchFromCerberus(Cerberus.java:59)
	at com.lookout.cerberus.client.Cerberus.retrieve(Cerberus.java:53)
	at com.lookout.bac.instruction.processor.FileSystemSecretSaver.process(FileSystemSecretSaver.java:30)
	at com.lookout.bac.instruction.processor.InstructionProcessor.processInstruction(InstructionProcessor.java:13)
	at com.lookout.bac.instruction.InstructionManager.process(InstructionManager.java:33)
	at com.lookout.bac.instruction.InstructionManager.processInstruction(InstructionManager.java:38)
	at com.lookout.bac.Application.main(Application.java:30)

See attached screenshots as well.

fieldju:
@dylan-raithel-lookout we will check this out.

PR #184, opened 2019-01-24
fix: Trim whitespace when restoring data that is potentially created …

fieldju:
fix: Trim whitespace when restoring data that is potentially created before we removed whitespace from the db via an upgrade script, and add another upgrade script to remove any whitespace that snuck in.

Fixes: https://github.com/Nike-Inc/cerberus-management-service/issues/183


fieldju:
@dylan-raithel-lookout please UAT the following release https://github.com/Nike-Inc/cerberus-management-service/releases/tag/v3.25.3-RC and let us know if it fixes: https://github.com/Nike-Inc/cerberus-management-service/issues/183

PR #186, opened 2019-02-08
Prevent sdbs with the same slug from being created

fieldju:


src/main/resources/com/nike/cerberus/migration/V1.6.5.0__add_sdb_name_slug.sql
```diff
@@ -0,0 +1,13 @@
+ALTER TABLE SAFE_DEPOSIT_BOX
```
fieldju: image

PR #189, opened 2019-03-06
* update database operation related dependencies

mayitbeegh:
* skipping 3.25.6 because release version is messed up


fieldju:
I want to see this branch deployed to an environment that uses Aurora and see the API tests pass.

PR #195, opened 2019-06-04
Update Gradle Config

fieldju:
- Updated Gradle to Gradle 5
- Replaced findBugs with spotBugs
- Updated various Gradle plugins to be compatible 
- Added security dependency checker plugin from OWASP
- Fixed spotBugs found bug

The plugin now creates the following report

image

The plugin is not set up to fail the build yet.

Once we do the dependency update ticket, we can enable that.


src/main/java/com/nike/cerberus/service/DistributedLockService.java
```diff
@@ -136,10 +136,13 @@ public void executeServerShutdownHook(ServerConfig serverConfig, Channel channel
         locks.forEach((name, lock) -> {
             try {
                 lock.release();
-                lock.semaphore.tryAcquire(3, TimeUnit.SECONDS);
+                boolean didAcquireLock = lock.semaphore.tryAcquire(3, TimeUnit.SECONDS);
```
mayitbeegh: Is this the bug SpotBugs found?

fieldju: Yeah, and the bug was real, basically I had assumed the interrupt exception was thrown when the timeout occurred, but in reality it returns false, which was being ignored.

PR #196, opened 2019-06-19
Feature/data key rotation

mayitbeegh:


src/main/java/com/nike/cerberus/service/SecureDataService.java
```diff
@@ -444,4 +446,93 @@ boolean secureDataHasBeenUpdated(SecureDataRecord secureDataRecord) {
     public int getTotalNumberOfFiles() {
         return secureDataDao.countByType(SecureDataType.FILE);
     }
+
+    public void rotateDataKeys(int numberOfKeys, int pauseTimeInMillis, int rotationIntervalInDays) {
```
mayitbeegh: Need some comments on the latest version first implementation

PR #199, opened 2019-08-06
refactor: Made Riposte endpoint be annotation based.

fieldju:
I am doing a POC to Springify our Riposte plus Guice setup so that we can eventually layer in Nike specific code via internal modules that get loaded via class path scanning.

### DIFF
--- a/src/main/java/com/nike/cerberus/endpoints/RiposteEndpoint.java
+++ b/src/main/java/com/nike/cerberus/endpoints/RiposteEndpoint.java
@@ -1,0 +1,14 @@
+package com.nike.cerberus.endpoints;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * An annotation that can be used to mark classes as Riposte endpoints and have them bootstrapped into the Guice module.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface RiposteEndpoint {
+}
--- a/src/main/java/com/nike/cerberus/server/config/guice/CmsGuiceModule.java
+++ b/src/main/java/com/nike/cerberus/server/config/guice/CmsGuiceModule.java
@@ -18,45 +18,14 @@
 package com.nike.cerberus.server.config.guice;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.google.inject.AbstractModule;
-import com.google.inject.Injector;
-import com.google.inject.Key;
-import com.google.inject.Provides;
+import com.google.inject.*;
 import com.google.inject.name.Names;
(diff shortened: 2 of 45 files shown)


src/main/java/com/nike/cerberus/server/config/guice/CmsGuiceModule.java
```diff
@@ -164,68 +130,12 @@ private void configureAuditLogging() {
...
-                deleteSafeDepositBox, updateSafeDepositBoxV1, updateSafeDepositBoxV2, createSafeDepositBoxV1, createSafeDepositBoxV2,
-                getSDBMetadata, putSDBMetadata, overrideSdbOwner,
-                writeSecureData, readSecureData, deleteSecureData,
-                triggerScheduledJob,
-                getDashboard, getDashboardRedirect,
-                writeSecureFile, readSecureFile, deleteSecureFile, headSecureFile, getSecureFiles,
-                restoreSafeDepositBox,
-                getSecretVersionPathsForSdb, getSecureDataVersions, getAuthKmsKeyMetadata
-        ));
+    public Set<Endpoint<?>> appEndpoints(Injector injector) {
```
fieldju: Now we can in theory not have to manually add endpoints to this and add custom endpoints in an internal module.

For example Tim wants us to be able to map user groups to individuals, which is likely to be Nike specific.

src/main/java/com/nike/cerberus/endpoints/RiposteEndpoint.java
```diff
@@ -0,0 +1,14 @@
...
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * An annotation that can be used to mark classes as Riposte endpoints and have them bootstrapped into the Guice module.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface RiposteEndpoint {
```
fieldju: Any Riposte endpoints can just be annotated with this now.

src/test/java/com/nike/cerberus/backstopper/apierror/contract/jsr303convention/ApplicationJsr303AnnotationTroller.java
```diff
@@ -1,64 +0,0 @@
-/*
```
mayitbeegh: Why are these deleted?

fieldju: Because they are doing weird things with all the annotations despite claiming that they are only for JSR 308 Javax Validation annotations and it was causing issues. These tests were from the original boilerplate code and are now getting in the way.

We can make a todo to bring them back if you feel strongly.

fieldju: Was dep conflict, fixed and brought back tests.

PR #202, opened 2019-09-13
Feature/jwt auth

mayitbeegh:


src/main/java/com/nike/cerberus/service/JwtService.java
```diff
@@ -0,0 +1,127 @@
...
+     * @return JWT token
+     */
+    public String generateJwtToken(CerberusJwtClaims cerberusJwtClaims){
+        CerberusJwtKeySpec cerberusJwtKeySpec = signingKeyResolver.resolveSigningKey();
+        String jwtToken = Jwts.builder()
+                .setHeaderParam(JwsHeader.KEY_ID, cerberusJwtKeySpec.getKid())
+                .setId(cerberusJwtClaims.getId())
+                .setIssuer(environmentName)
+                .setSubject(cerberusJwtClaims.getPrincipal())
+                .claim(PRINCIPAL_TYPE_CLAIM_NAME, cerberusJwtClaims.getPrincipalType())
```
fieldju: I think you might consider making an object for the claims and use Jackson to manage serializing and deserializing it to and from a map rather than manually doing it.

src/test/java/com/nike/cerberus/service/JwtServiceTest.java
```diff
@@ -0,0 +1,92 @@
...
+import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
+import java.util.Optional;
+
+import static org.junit.Assert.*;
+import static org.mockito.Matchers.*;
+import static org.mockito.Mockito.when;
+import static org.mockito.MockitoAnnotations.initMocks;
+
+public class JwtServiceTest {
```
fieldju: I would like to see more tests here.

Specifically a test proving that no-alg as the alg doesn't circumvent auth.

fieldju:
I don't see where you are providing the default job configuration for the synchronizer job:
https://github.com/Nike-Inc/cerberus-management-service/blob/master/src/main/resources/cms.conf#L122-L125

- [ ] Add default job config

PR #203, opened 2019-09-16
Enable data key caching in AWS crypto operations

mayitbeegh:


src/main/resources/cms.conf
```diff
@@ -196,3 +196,10 @@ cms.iam.token.ttl=1h
 #   When true, if an SDB grants access to AD group 'Lst-foo', then users in group 'Lst-Foo' will not have access
 #   When false, if an SDB grants access to AD group 'Lst-foo', then users in group 'Lst-Foo' will have access
 cms.user.groups.caseSensitive=true
+
+cms.encryption.encrypt.cache.maxSize=0
+cms.encryption.encrypt.cache.maxAge=5
```
fieldju: I think we were going to have it default to 0 here, basically disabling caching and then manually enable it via the Cerberus Lifecycle cli.

src/main/java/com/nike/cerberus/server/config/guice/CmsGuiceModule.java
```diff
@@ -287,4 +298,52 @@ public StaticAssetManager dashboardStaticAssetManager() {
...
+
+    @Provides
+    @Singleton
+    @Named("encryptCryptoMaterialsManager")
+    public CryptoMaterialsManager encryptCryptoMaterialsManager(@Named("cms.encryption.cmk.arns") String cmkArns,
+                                                                @Named("cms.encryption.encrypt.cache.maxSize") int maxSize,
+                                                                @Named("cms.encryption.encrypt.cache.maxAge") int maxAge,
+                                                                @Named("cms.encryption.encrypt.cache.messageUseLimit") int messageUseLimit,
+                                                                Region currentRegion) {
+        MasterKeyProvider<KmsMasterKey> keyProvider = initializeKeyProvider(cmkArns, currentRegion);
```
fieldju: Can you add a log statement here so we can see what the conf is at boot time.

src/main/java/com/nike/cerberus/server/config/guice/CmsGuiceModule.java
```diff
@@ -287,4 +298,52 @@ public StaticAssetManager dashboardStaticAssetManager() {
...
+        return cachingCmm;
+    }
+
+    @Provides
+    @Singleton
+    @Named("decryptCryptoMaterialsManager")
+    public CryptoMaterialsManager decryptCryptoMaterialsManager(@Named("cms.encryption.cmk.arns") String cmkArns,
+                                                                @Named("cms.encryption.decrypt.cache.maxSize") int maxSize,
+                                                                @Named("cms.encryption.decrypt.cache.maxAge") int maxAge,
+                                                                Region currentRegion) {
```
fieldju: Can you add a log statement here so we can see what the conf is at boot time.

gradle.properties
```diff
@@ -14,6 +14,6 @@
 # limitations under the License.
 #
 
-version=3.30.0
+version=3.32.0
```
fieldju: Probably revert this and just change it manually, I think this pr will might merged before the other one.

src/main/resources/cms.conf
mayitbeegh: I set it to 5 because the constructor doesn't like 0. I can throw an if statement in front of it but since the cache size is set to 0 there will be no caching either way

gradle.properties
mayitbeegh: 32 is the correct next version because the last release 3.31.0 didn't come with gradle.properties update. The other PR will be 3.33.0

src/main/java/com/nike/cerberus/server/config/guice/CmsGuiceModule.java
mayitbeegh: Done

src/main/java/com/nike/cerberus/server/config/guice/CmsGuiceModule.java
mayitbeegh: Done

src/main/java/com/nike/cerberus/server/config/guice/CmsGuiceModule.java
```diff
@@ -287,4 +298,56 @@ public StaticAssetManager dashboardStaticAssetManager() {
...
+        logger.info("Initializing encryptCryptoMaterialsManager with CMK: {}, maxSize: {}, maxAge: {}, " +
+                "messageUseLimit: {}", cmkArns, maxSize, maxAge, messageUseLimit);
+        MasterKeyProvider<KmsMasterKey> keyProvider = initializeKeyProvider(cmkArns, currentRegion);
+        CryptoMaterialsCache cache = new LocalCryptoMaterialsCache(maxSize);
+        CryptoMaterialsManager cachingCmm =
+                CachingCryptoMaterialsManager.newBuilder().withMasterKeyProvider(keyProvider)
+                        .withCache(cache)
+                        .withMaxAge(maxAge, TimeUnit.SECONDS)
+                        .withMessageUseLimit(messageUseLimit)
```
tlisonbee: You don't really need any caching or re-use on write since writes are so infrequent.  If you are going to spend memory on a cache it seems better to spend zero on write and spend all of it on read.

fieldju: We generally agree with you, but with we figured we would like a tiny cache here just incase a user goes off the rails and does something wrong so that we don't blow through our KMS API limit and inadvertently get DDOS'd by a user. 

With the mandated move away from AWS WAF to our internal WAF we are going to lose our current rate limiter unless we re-engineer it to be a 1st class thing in the app.

Under normal usages, we won't really end up using the cache because they are so infrequent, so it's really there just in-case some sort of burst traffic happens due to a user doing something wrong.

PR #205, opened 2019-09-25
Feature/jwt auth blacklist

mayitbeegh:


src/main/java/com/nike/cerberus/jobs/JwtBlacklistCleanUpJob.java
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright (c) 2017 Nike, Inc.
```
fieldju: - [ ] Fix date

fieldju:
I don't see where you are providing the default job configuration for the synchronizer job:
https://github.com/Nike-Inc/cerberus-management-service/blob/master/src/main/resources/cms.conf#L122-L125

- [ ] Add default job config

PR #211, opened 2019-10-01
Add caching to KMS auth

mayitbeegh:


src/main/resources/cms.conf
```diff
@@ -198,3 +198,4 @@ cms.iam.token.ttl=1h
 cms.user.groups.caseSensitive=true
 
 cms.encryption.cache.enabled=false
+cms.iam.token.cache.enable=false
```
fieldju: Can you rename to `cms.auth.iam.kms.cache.enabled`

mayitbeegh: There's another config `cms.iam.token.ttl` that's somewhat related. But I can change it.

PR #214, opened 2019-10-18
refactor: Removed all legacy policy generation logic that hasn't been…

fieldju:
… necessary since Highlander (implenting /v1/secret internally). Policies are no longer used by this service, so we will just return a empty list.


src/test/java/com/nike/cerberus/service/AuthenticationServiceTest.java
```diff
@@ -489,4 +440,4 @@ public void tests_that_refreshUserToken_throws_access_denied_token_when_count_is
         assertTrue(e instanceof ApiException);
         assertTrue(((ApiException) e).getApiErrors().contains(DefaultApiError.MAXIMUM_TOKEN_REFRESH_COUNT_REACHED));
     }
-}
\ No newline at end of file
+}
```
mayitbeegh: Outrageous

PR #225, opened 2020-01-03
Feature/configure athena audit logging

melanahammel:


settings.gradle
```diff
@@ -27,3 +28,4 @@ def setBuildFile(project) {
 rootProject.children.each {
   setBuildFile(it)
 }
+include 'cerberus-audit-logger-athena'
```
fieldju: why is this line here 2x

cerberus-web/src/main/java/com/nike/cerberus/event/listener/AuditLoggingEventListener.java
```diff
@@ -24,8 +24,8 @@
 
 /** Event listener implementation that logs each event. */
 @Slf4j
+@ConditionalOnProperty("cerberus.events.auditLoggingEventListener.enabled")
```
fieldju: This should be in the config class that then optionally scans everything in `com.nike.cerberus.audit.athena`
The config class should be in `com.nike.cerberus.config`

fieldju:
Can you port this too: https://github.com/Nike-Inc/cerberus/blob/master/src/test/java/com/nike/cerberus/service/S3LogUploaderServiceTest.java

settings.gradle
melanahammel: - [x] remove duplicate line

cerberus-web/src/main/java/com/nike/cerberus/config/ApplicationConfiguration.java
```diff
@@ -48,6 +50,7 @@
 @ComponentScan({
   "com.netflix.spinnaker.kork.secrets",
   "com.nike.backstopper", // error management TODO explicitly import the config bean
+  "com.nike.cerberus.audit.logger",
```
fieldju: this should be removed and handled in your configuration class.

cerberus-web/src/main/java/com/nike/cerberus/event/listener/AuditLoggingEventListener.java
melanahammel: - [x] move @ConditionalOnProperty to AthenaAuditLoggerConfiguration

melanahammel:
- [x] port S3LogUploaderServiceTest.java from old project

cerberus-web/src/main/java/com/nike/cerberus/config/ApplicationConfiguration.java
fieldju: example, your config class is in com.nike.cerberus.config, which is already loaded.

then in your config class you have something like the following 

```java
@Configuration
@ConditionalOnProperty("cerberus.userAuth.connector.okta.enabled")
@ComponentScan({"com.nike.cerberus.auth.connector.okta"})
```

melanahammel: - [x] move @ComponentScan for logger to AthenaAuditLoggerConfiguration

PR #238, opened 2020-02-03
Cerberus (Phoenix) 4.0.0-RC1

fieldju:
**TL;DR**: We refactored Cerberus from a Guice/Riposte project to a composable multi-project Springboot application.

**The changes should be mostly backward compatible**

We did technically make a few "_Breaking changes_" that you shouldn't see on the normal happy path.

0. When you list secrets on an empty SDB / Path an empty list is now returned instead of a 404.
Most of our clients already catch the 404 and make it an empty list anyways.
This was legacy behavior from when we still used Hashicorp Vault, which just confused people, so we fixed it.
1. Some of the status codes changed from 200 to 201 or 204 and vice versa to be more standardized as Springboot enforces some of this behavior by default.
2.  Some of the status codes changed from 400 to 401 or 403 and vice versa to be more standardized as Springboot enforces some of this behavior by default.
3. We now always return a Backstopper error object, previously on /v1/secret/*** we returned a legacy Vault error object.
This might break the unhappy path on some of our clients and cause a deserialization issue to bubble up instead of the real underlying cause.
We hypothesis that this will only affect new uses of Cerberus and should have no impact on projects already established with Cerberus.
We are working on updating any clients that don't handle this change gracefully. 

**Additional Notes**

This is going to put the create an environment docs and life cycle management in a bad state while we work on updating those.


API.md
```diff
@@ -1,1635 +0,0 @@
-# Cerberus Management Service
```
fieldju: - [x] Revert this deletion

melanahammel (approved):

PR #241, opened 2020-02-13
fix: corrected file path for migrations

melanahammel:


mayitbeegh (approved):

fieldju:
I had to do this when I ran this locally the other day, because it wasn't running the migration scripts on a new db. I forgot to make PR :(.

fieldju:
image

I had mine end with a / not sure if it maters though

melanahammel:
In these [docs](https://docs.spring.io/spring-boot/docs/current/reference/html/appendix-application-properties.html#data-migration-properties) if you go down to `spring.flyway.locations`, the example classpath does not end with a slash. It seems like it works both ways.

PR #254, opened 2020-03-16
fix: redact sensitive header in error message

mayitbeegh:


rdong333 (approved):

cerberus-web/src/main/java/com/nike/cerberus/error/SfxAwareApiExceptionHandlerUtils.java
```diff
@@ -81,4 +89,15 @@ public String buildErrorMessageForLogs(
           .inc();
     }
   }
+
+  private void reactSensitiveHeaders(RequestInfoForLogging request) {
```
melanahammel: Should `reactSensitiveHeaders` be `redactSensitiveHeaders`?

melanahammel (changes requested):

cerberus-web/src/main/java/com/nike/cerberus/error/SfxAwareApiExceptionHandlerUtils.java
mayitbeegh: Doh

melanahammel (approved):

PR #263, opened 2020-08-03
Build/update owasp plugin and fix dependencies

tunderwood:
Updated the OWASP dependency plugin to version 5.3.2 and took care of the resulting dependencies that were found to be vulnerable.


cerberus-dashboard/package.json
```diff
@@ -12,7 +12,8 @@
     "start": "react-scripts start",
     "build": "react-scripts build && rm -rf dist && mv build dist",
     "test": "react-scripts test --env=jsdom",
-    "logs": "multitail logs/*"
+    "logs": "multitail logs/*",
+    "preinstall": "npx npm-force-resolutions"
```
mayitbeegh, line 16: what's this?

dependency-check-supressions.xml
```diff
@@ -83,4 +83,11 @@
         <packageUrl regex="true">^pkg:maven/org\.dom4j/dom4j@.*$</packageUrl>
         <cve>CVE-2020-10683</cve>
     </suppress>
+    <suppress>
+        <notes><![CDATA[
+            file name: websocket-extensions:0.1.4
+   ]]></notes>
+        <packageUrl regex="true">^pkg:npm/websocket\-extensions@.*$</packageUrl>
+        <vulnerabilityName>CVE-2020-7663</vulnerabilityName>
```
mayitbeegh, line 91: Why is this suppressed?

tunderwood: the most recent version of websocket-extensions (0.1.4) has a vulnerability in it that has not yet been patched. Until a fix is released vulnerability needs to be suppressed for the build to pass.

cerberus-dashboard/package.json
tunderwood: npm-force-resolutions is an npm plugin that forces your node dependencies to a specific version. 

In this case, react-scripts@3.4.1 (the most current version of react-scripts) pulls in yargs-parser 11.1.1 which has a high severity CVE. We need to use npm-force-resolutions to force yargs-parser to 13.1.2 where the vulnerability has been resolved. Once react-scripts updates their version of yargs-parser we can unforce this. 

Here are the package details. https://github.com/rogeriochaves/npm-force-resolutions

dependency-check-supressions.xml
tunderwood: It should be noted that react-scripts is what bring in websocket-extensions.React-scripts is a devDependency, not a production dependency so this shouldn't expose production instances to increased vulnerabilities.

mayitbeegh (approved):

melanahammel (approved):

PR #265, opened 2020-08-07
Feature/aws china

mayitbeegh:


mayitbeegh:
The (aws|aws-cn) portion is call "partition" in this AWS doc
https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html

cerberus-web/src/main/java/com/nike/cerberus/aws/sts/AwsStsHttpClient.java
```diff
@@ -115,9 +116,15 @@ public AwsStsHttpClient(OkHttpClient httpClient, ObjectMapper objectMapper) {
 
   /** Build the request */
   protected Request buildRequest(String region, Map<String, String> headers) {
+    String stsEndpointUrl;
+    if (region.startsWith("cn-")) {
```
tunderwood, line 120: Do we think there is ever a chance that a new region is added to the China partition that does not begin with CN? Would it be safer to have a list of the regions (or region prefixes) to compare against to make this easier to update in the future?

cerberus-web/src/main/java/com/nike/cerberus/service/AuthenticationService.java
```diff
@@ -221,7 +221,10 @@ public EncryptedAuthDataWrapper authenticate(IamRoleCredentials credentials) {
 
     final String iamPrincipalArn =
         String.format(
-            AWS_IAM_ROLE_ARN_TEMPLATE, credentials.getAccountId(), credentials.getRoleName());
+            AWS_IAM_ROLE_ARN_TEMPLATE,
+            "aws",
```
tunderwood: We are hardcoding this to the AWS global partition, is there ever a chance it will need to be aws-cn?

cerberus-web/src/main/java/com/nike/cerberus/service/SafeDepositBoxService.java
```diff
@@ -636,6 +647,7 @@ protected SafeDepositBoxV2 convertSafeDepositBoxV1ToV2(SafeDepositBoxV1 safeDepo
                         .withIamPrincipalArn(
                             String.format(
                                 DomainConstants.AWS_IAM_ROLE_ARN_TEMPLATE,
+                                "aws",
```
tunderwood: Is there ever a chance this will need to be the China partition?

cerberus-web/src/main/java/com/nike/cerberus/util/AwsIamRoleArnParser.java
```diff
@@ -18,14 +18,25 @@
...
 import org.springframework.stereotype.Component;
 
 /** Utility class for concatenating and parsing AWS IAM role ARNs. */
 @Component
 public class AwsIamRoleArnParser {
+  private final boolean awsChinaAllowed;
+  private final boolean awsGlobalAllowed;
+
+  public AwsIamRoleArnParser(
+      @Value("${cerberus.partitions.awsGlobal.enabled}") boolean awsGlobalAllowed,
```
tunderwood: nit: Do we care that in config this is "enabled" vs in code it's "allowed"?

cerberus-web/src/test/java/com/nike/cerberus/service/AuthenticationServiceTest.java
```diff
@@ -266,7 +266,7 @@ public void test_that_getKeyId_only_validates_kms_policy_one_time_within_interva
     String accountId = "0000000000";
     String roleName = "role/path";
     String principalArn = String.format("arn:aws:iam::%s:instance-profile/%s", accountId, roleName);
-    String roleArn = String.format(AWS_IAM_ROLE_ARN_TEMPLATE, accountId, roleName);
+    String roleArn = String.format(AWS_IAM_ROLE_ARN_TEMPLATE, "aws", accountId, roleName);
```
tunderwood: Should we use a constant for partition vs a hardcoded string? Do we need to do any testing for "aws-cn"?

cerberus-web/src/test/java/com/nike/cerberus/service/AuthenticationServiceTest.java
```diff
@@ -311,7 +311,7 @@ public void test_that_findIamRoleAssociatedWithSdb_returns_empty_optional_when_r
...
```
tunderwood: Should we use a constant for partition vs a hardcoded string? Do we need to do any testing for "aws-cn"?

cerberus-web/src/main/java/com/nike/cerberus/aws/sts/AwsStsHttpClient.java
mayitbeegh: Either way the code needs to be reevaluated whenever a new AWS China region comes out. If we use a list then we need to update the list every time. I'd argue that this is a pretty good assumption about the prefix (until it's not).

cerberus-web/src/main/java/com/nike/cerberus/service/AuthenticationService.java
mayitbeegh: There is a chance. Right now it's only there for backwards compatibility. Maybe we can use aws/aws-cn depending on the feature flag or adding an extra field in the V1 KMS controller. But since the KMS is deprecated, I'd avoid putting too much working into it.

cerberus-web/src/main/java/com/nike/cerberus/service/SafeDepositBoxService.java
mayitbeegh: This is also part of a deprecated API. Same reasoning as above.

> There is a chance. Right now it's only there for backwards compatibility. Maybe we can use aws/aws-cn depending on the feature flag or adding an extra field in the V1 KMS controller. But since the KMS is deprecated, I'd avoid putting too much working into it.

cerberus-web/src/main/java/com/nike/cerberus/util/AwsIamRoleArnParser.java
mayitbeegh: Good catch. I'll change it to "enabled".

cerberus-web/src/test/java/com/nike/cerberus/service/AuthenticationServiceTest.java
mayitbeegh: I'm debating on it. The constant hurts the readability of the regex imo if we decide to replace all occurrences of the partition name. But maybe we can just use constant outside of regex. Thoughts?

cerberus-web/src/test/java/com/nike/cerberus/service/AuthenticationServiceTest.java
mayitbeegh: I'll add/modify some tests for it

cerberus-web/src/main/java/com/nike/cerberus/service/SafeDepositBoxService.java
```diff
@@ -309,6 +314,12 @@ public SafeDepositBoxV2 updateSafeDepositBoxV2(
     final Set<IamPrincipalPermission> iamRolePermissionSet =
         safeDepositBox.getIamPrincipalPermissions();
 
+    iamRolePermissionSet.stream()
```
shawn-sher: This section of code is identical to the code above starting on line 249, maybe it could be pulled into a method?

mayitbeegh: Sure

cerberus-web/src/test/java/com/nike/cerberus/service/AuthenticationServiceTest.java
mayitbeegh: Actually I don't think the tests here validate or invalidate this change. Most of the stuffs that matter are mocked. I threw a test in there anyway.

shawn-sher (approved):

cerberus-web/src/main/java/com/nike/cerberus/util/AwsIamRoleArnParser.java
```diff
@@ -18,14 +18,25 @@
...
 import org.springframework.stereotype.Component;
 
 /** Utility class for concatenating and parsing AWS IAM role ARNs. */
 @Component
 public class AwsIamRoleArnParser {
+  private final boolean awsChinaEnabled;
+  private final boolean awsGlobalEnabled;
+
+  public AwsIamRoleArnParser(
+      @Value("${cerberus.partitions.awsGlobal.enabled}") boolean awsGlobalEnabled,
```
melanahammel, line 35: Do we want both AWS Global and AWS China to be able to be enabled at the same time? Do we want to be more prescriptive and block that functionality for security reasons?

tunderwood: I don't think it hurts to allow both since it may be a valid use case for some organizations.

cerberus-web/src/test/java/com/nike/cerberus/service/AuthenticationServiceTest.java
tunderwood: I'm a pretty big fan of consistency, since we use variables for accountId and roleName I think it makes sense to do the same for the partition. I agree about the readability of the regex but I don't think it makes it any worse than what is already there.

tunderwood (approved):

PR #273, opened 2020-10-01
Feat/enable okta push notification

tunderwood:
Upgraded the Okta SDKs and added support for Okta Verify's push notification multi-factor authentication.


cerberus-auth-connector-okta/src/main/java/com/nike/cerberus/auth/connector/okta/OktaAuthConnector.java
```diff
@@ -96,6 +99,36 @@ public AuthResponse triggerChallenge(String stateToken, String deviceId) {
...
+
+    CompletableFuture<AuthResponse> authResponseFuture = new CompletableFuture<>();
+    PushStateHandler stateHandler =
+        new PushStateHandler(oktaAuthenticationClient, authResponseFuture);
+
+    try {
+      oktaAuthenticationClient.verifyFactor(deviceId, stateToken, stateHandler);
+
+      AuthResponse authResponse = authResponseFuture.get(45, TimeUnit.SECONDS);
+      while (authResponse.getData().getFactorResult().equals("WAITING")) {
```
melanahammel: How long will this while loop last? Meaning how long is a push notification good for until it expires?

gradle.properties
```diff
@@ -16,4 +16,4 @@
 
 version=4.8.1
```
melanahammel: Don't forget to bump version

cerberus-dashboard/src/reducers/authenticationReducer.js
```diff
@@ -108,7 +108,11 @@ export default createReducer(initialState, {
         return Object.assign({}, state, {
             selectedDeviceId: payload.selectedDeviceId,
             shouldDisplaySendCodeButton: state.mfaDevices
-                .filter(device => device.id === payload.selectedDeviceId)[0].requires_trigger
+                .filter(device => device.id === payload.selectedDeviceId)[0].requires_trigger &&
+                !state.mfaDevices.filter(device => device.id === payload.selectedDeviceId)[0].is_push,
+            shouldDisplaySendPushButton: state.mfaDevices
```
melanahammel, line 112: Trying to parse this code here. Is this setting the default option for users as a push notification if it is enabled?

cerberus-dashboard/src/actions/authenticationActions.js
```diff
@@ -245,7 +245,44 @@ export function triggerCodeChallenge(mfaDeviceId, stateToken) {
...
+                device_id: mfaDeviceId,
+                is_push: true
+            },
+            timeout: AUTH_ACTION_TIMEOUT
+        })
+            .then(function (response) {
+                handleUserLogin(response, dispatch);
+            })
+            .catch(function ({ response }) {
+                log.error('Failed to finalize MFA login', response);
```
melanahammel, line 275: I tested this, and response is undefined when I let the push notification time out, not sure if that the expected behavior or not.

A better error message might be helpful here, maybe saying something about how it could be due to timeout. It threw a 504, which implies server error, but it was because I didn't click on the push notification which could be seen as a client error.

cerberus-dashboard/src/actions/authenticationActions.js
```diff
@@ -245,7 +245,44 @@ export function triggerCodeChallenge(mfaDeviceId, stateToken) {
...
+        dispatch(loginMfaChallenge());
+        return axios({
+            method: 'post',
+            url: environmentService.getDomain() + cms.USER_AUTH_MFA_PATH,
+            data: {
+                state_token: stateToken,
+                device_id: mfaDeviceId,
+                is_push: true
+            },
+            timeout: AUTH_ACTION_TIMEOUT
```
melanahammel, line 269: This is 60s. Is that standard for push notifications? Does that give users enough time?

cerberus-auth-connector-okta/src/main/java/com/nike/cerberus/auth/connector/okta/OktaAuthConnector.java
melanahammel: After reading the docs, it looks like Okta will return `TIMEOUT`, so the while loop then would exit.

https://developer.okta.com/docs/reference/api/factors/#factor-verify-result-object

cerberus-dashboard/src/reducers/authenticationReducer.js
tunderwood: Let me take a look at this as it may no longer be needed. I did some refactoring to this workflow later into my changes. I'll update when I confirm.

cerberus-dashboard/src/actions/authenticationActions.js
tunderwood: I kept the same timeout we use for OTP, SMS, and phone. In my experience, 60s should be more than enough time. I haven't timed it but I feel like push MFA is the fastest of the options (no need to find and open an app, wait for a text, or wait for a phone call, and then type the code ). I don't know if there is a standard amount but my thought was if it's long enough for the other factor types it should be long enough for push style.

cerberus-dashboard/src/actions/authenticationActions.js
tunderwood: good catch, I'll see about adding some better error handling here.

cerberus-auth-connector-okta/src/main/java/com/nike/cerberus/auth/connector/okta/OktaAuthConnector.java
tunderwood: Yeah, one thing I'm still not clear on is how long they wait. The UI will wait 60s prior to timing out which I feel is a good amount of time, not sure how long it takes for the okta factor to expire though. Worth exploring so we can ensure we error appropriately.

gradle.properties
tunderwood: 👍

cerberus-dashboard/src/actions/authenticationActions.js
melanahammel: sounds reasonable to me

cerberus-dashboard/src/reducers/authenticationReducer.js
tunderwood: Ok, I just reviewed this. 

Essentially what this does is determine the layout of the page. Originally I was reusing the "requires_trigger" parameter to help determine this but I refactored that for clarity. My next update will clean this up.

cerberus-auth-connector-okta/src/main/java/com/nike/cerberus/auth/connector/okta/OktaAuthConnector.java
tunderwood: I left it at 60 seconds for time out. The Okta timeout is over 5 minutes so I just don't think it makes sense to wait that long. I also updated the system to let the client know about the lack of response so the errors inform the user.

cerberus-dashboard/src/actions/authenticationActions.js
tunderwood: Did a lot of updates to errors. should now give an unauthorized error code with a timeout and inform the user of why it happened.

cerberus-web/src/main/java/com/nike/cerberus/service/AuthenticationService.java
```diff
@@ -186,6 +186,30 @@ public AuthResponse triggerChallenge(final MfaCheckRequest challengeRequest) {
...
+            challengeRequest.getStateToken(), challengeRequest.getDeviceId());
+    if (authResponse.getStatus() == AuthStatus.SUCCESS) {
+      authResponse
+          .getData()
+          .setClientToken(
+              generateToken(
+                  authResponse.getData().getUsername(),
+                  authServiceConnector.getGroups(authResponse.getData()),
+                  0));
+    } else {
```
melanahammel, line 207: Should this else block be empty?

melanahammel: Should this else block be empty?

melanahammel (changes requested):

cerberus-auth-connector-okta/src/main/java/com/nike/cerberus/auth/connector/okta/OktaAuthConnector.java
```diff
@@ -96,6 +99,52 @@ public AuthResponse triggerChallenge(String stateToken, String deviceId) {
...
+  public AuthResponse triggerPush(String stateToken, String deviceId) {
+
+    CompletableFuture<AuthResponse> authResponseFuture = new CompletableFuture<>();
+    PushStateHandler stateHandler =
+        new PushStateHandler(oktaAuthenticationClient, authResponseFuture);
+
+    try {
+      oktaAuthenticationClient.verifyFactor(deviceId, stateToken, stateHandler);
+
+      AuthResponse authResponse = authResponseFuture.get(45, TimeUnit.SECONDS);
```
shawn-sher, line 112: What's 45?

cerberus-web/src/main/java/com/nike/cerberus/service/AuthenticationService.java
tunderwood: Nope, that got caught from a copy past. Nice catch.

melanahammel (approved):

PR #285, opened 2020-12-08
Dep/dependency upgrade staging

tunderwood:


build.gradle
```diff
@@ -64,21 +66,42 @@ allprojects {
...
+        destination = file("$buildDir/reports/spotbugs/main/spotbugs.html")
+        stylesheet = 'fancy-hist.xsl'
+      }
+      xml {
+        enabled = false
+      }
+    } as Closure<NamedDomainObjectContainer<? extends SpotBugsReport>>))
   }
 
+  spotbugsTest {
```
melanahammel, line 89: Does this look for bugs in test files? Or is this a test of SpotBugs?

cerberus-web/build.gradle
```diff
@@ -15,7 +15,7 @@
  */
 
 plugins {
-  id 'org.springframework.boot' version "${springBootVersion}"
+  id 'org.springframework.boot' version "2.4.0"
```
melanahammel, line 18: I think we should keep this to `"${springBootVersion}"` so that way it stays in sync with the versioning in `build.gradle` at the root.

build.gradle
tunderwood: Bugs in the Test files

cerberus-web/build.gradle
tunderwood: I don't disagree, Dependabot didn't do the spring upgrade when it was set that way which is why I made the change. If we are fine excluding Spring from automatic upgrades we can switch it back but it's probably worth having a discussion.

melanahammel: Okay sounds good, let's discuss it tomorrow

melanahammel (approved):

PR #445, opened 2021-03-31
Improving unit test coverage for Cerberus server

RVelp1:


cerberus-audit-logger-athena/src/test/java/com/nike/cerberus/audit/logger/service/AthenaServiceTest.java
```diff
@@ -0,0 +1,40 @@
...
+
+  @Test
+  public void test_that_addPartition_works() {
+    String fileName = "localhost-audit.2018-01-29_12-58.log.gz";
+    try {
+      athenaService.addPartitionIfMissing("us-west-2", "fake-bucket", "2018", "01", "29", "12");
+    } catch (Exception e) {
+      assertTrue(false);
+    }
+    assertTrue(true);
```
tbishopric: Wait, what does this do? You are asserting that true == true? What is this testing?

cerberus-audit-logger-athena/src/test/java/com/nike/cerberus/audit/logger/service/AthenaServiceTest.java
```diff
@@ -0,0 +1,40 @@
...
+import static org.mockito.MockitoAnnotations.initMocks;
+
+import com.nike.cerberus.audit.logger.AthenaClientFactory;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mock;
+
+public class AthenaServiceTest {
+
+  AthenaClientFactory athenaClientFactory = new AthenaClientFactory();
```
tbishopric: How about putting initialization code in a constructor or the @Before method? What is the scope of this member variable (public, private, etc.)?

cerberus-audit-logger-athena/src/test/java/com/nike/cerberus/audit/logger/service/S3LogUploaderServiceTest.java
```diff
@@ -51,4 +53,35 @@ public void test_that_getPartition_works() {
...
+    assertEquals(expected, actual);
+  }
+
+  @Test
+  public void test_ingest_log_works() {
+    try {
+      String fileName = "localhost-audit.2018-01-29_12-58.log.gz";
+      // File f = new File(fileName);
+      s3LogUploader.ingestLog(fileName);
+      Thread.currentThread().sleep(TimeUnit.SECONDS.toMillis(60));
```
tbishopric: Why are you forcing a thread to sleep? This is generally not a good practice for unit tests. They should run as fast as possible.

RVelp1: Thread.sleep is used because the actual method being called in service  is submitted to a executor service
 And the actual test method thread gets terminated before the executor service executes the intended lines as the method inside the service is forced to sleep for 1sec 30times. To improve the coverage we need to sleep the main thread till the execution of executor service thread completess

cerberus-audit-logger-athena/src/test/java/com/nike/cerberus/audit/logger/service/AthenaServiceTest.java
RVelp1: The method being called should be executed smoothly and if at all an exception occurs it is  not expected. This is test for void method and actually not required a junit.

cerberus-audit-logger-athena/src/test/java/com/nike/cerberus/audit/logger/service/AthenaServiceTest.java
RVelp1: i will  modify this

cerberus-audit-logger-athena/src/test/java/com/nike/cerberus/audit/logger/service/AthenaServiceTest.java
RVelp1: i will remove the assert true, it's by mistake, i did not notice

cerberus-audit-logger-athena/src/test/java/com/nike/cerberus/audit/logger/service/AthenaServiceTest.java
```diff
@@ -0,0 +1,40 @@
...
+    athenaService = new AthenaService("fake-bucket", athenaClientFactory);
+  }
+
+  @Test
+  public void test_that_addPartition_works() {
+    String fileName = "localhost-audit.2018-01-29_12-58.log.gz";
+    try {
+      athenaService.addPartitionIfMissing("us-west-2", "fake-bucket", "2018", "01", "29", "12");
+    } catch (Exception e) {
+      assertTrue(false);
```
tbishopric: So is this assertion meant to make sure the method doesn't throw an Exception? You could just remove the try/catch. If it throws, then the exception will go up the stack and the test will fail. Otherwise you can use the fail() assertion instead of assertTrue(false).

cerberus-audit-logger-athena/src/test/java/com/nike/cerberus/audit/logger/service/S3LogUploaderServiceTest.java
tbishopric: I think we need to do something about this sleep. It's a really bad practice for unit tests. Are you trying to assert that we're using the executor service correctly? Or are you trying to test the actual log processing? I don't see this test asserting that any of the log processing actually happened. Maybe this test is not very useful.

cerberus-audit-logger-athena/src/test/java/com/nike/cerberus/audit/logger/service/S3LogUploaderServiceTest.java
```diff
@@ -51,4 +53,35 @@ public void test_that_getPartition_works() {
...
+      assertTrue(true);
+    } catch (Exception e) {
+      assertTrue(false);
+    }
+  }
+
+  @Test
+  public void test_executor_shutdown_works() {
+    try {
+      s3LogUploader.executeServerShutdownHook();
```
tbishopric: This test doesn't actually test that anything happens in that method. You're just asserting that it doesn't throw an exception. What about the actual logic in that method? Are we shutting down the executor correctly? Are we waiting for it to terminate cleanly? Etc.

cerberus-audit-logger-athena/src/test/java/com/nike/cerberus/audit/logger/service/AthenaServiceTest.java
tbishopric: And this test doesn't actually test any logic inside that addPartitionIfMissing() method. How much value is this test providing if it doesn't test and assert that the function behaves as written?

cerberus-audit-logger-athena/src/test/java/com/nike/cerberus/audit/logger/service/S3LogUploaderServiceTest.java
jiangha4: Instead of asserting false on assertTrue to test for an exception or not, you should have the test just catch the exception.

cerberus-audit-logger-athena/src/main/java/ch/qos/logback/core/rolling/AuditLogsS3TimeBasedRollingPolicy.java
```diff
@@ -65,4 +66,9 @@ public void rollover() throws RolloverFailure {
       }
     }
   }
+
+  void superRollOver() {
```
shawn-sher, line 70: What is this for?

RVelp1: this method is to mock super class rollover method

cerberus-audit-logger-athena/src/main/java/com/nike/cerberus/audit/logger/listener/AthenaLoggingEventListener.java
```diff
@@ -71,30 +71,20 @@ public void onApplicationEvent(AuditableEvent event) {
...
                       .put(
                           "principal_token_created",
                           cerberusPrincipal
-                              .map(
-                                  p ->
-                                      cerberusPrincipal
-                                          .get()
-                                          .getCreated()
-                                          .format(ATHENA_DATE_FORMATTER))
+                              .map(p -> p.getCreated().format(ATHENA_DATE_FORMATTER))
```
shawn-sher, line 79: These look more like fixes, are we sure they are equivalent?

RVelp1: yes they are equivalent

cerberus-audit-logger-athena/src/test/java/com/nike/cerberus/audit/logger/service/S3LogUploaderServiceTest.java
```diff
@@ -51,4 +52,40 @@ public void test_that_getPartition_works() {
...
+  }
+
+  @Test
+  public void test_ingest_log_works() {
+    String fileName = "localhost-audit.2018-01-29_12-58.log.gz";
+    s3LogUploader.ingestLog(fileName);
+    String fileName1 = "localhost-audit.2019-01-29_12-58.log.gz";
+    s3LogUploader.ingestLog(fileName);
+    s3LogUploader.ingestLog(fileName1);
+    //    try {
```
shawn-sher, line 71: Can we either uncomment or remove this test

shawn-sher: Can we either uncomment or remove this test?

cerberus-auth-connector-okta/src/test/java/com/nike/cerberus/auth/connector/okta/OktaApiClientHelperTest.java
```diff
@@ -54,6 +55,17 @@ public void setup() {
...
   // Test Methods
   /////////////////////////
+  @Test
+  public void oktaApiClientHelper() {
+    OktaConfigurationProperties oktaConfigurationProperties =
+        mock(OktaConfigurationProperties.class);
+    when(oktaConfigurationProperties.getBaseUrl()).thenReturn("url");
+    when(oktaConfigurationProperties.getApiKey()).thenReturn("api_key");
+    this.oktaApiClientHelper = new OktaApiClientHelper(oktaConfigurationProperties);
+    assertNotNull(oktaApiClientHelper);
```
shawn-sher: This looks like it only tests the mocks in this method and the constructor for OktaApiClientHelper.

cerberus-audit-logger-athena/src/test/java/com/nike/cerberus/audit/logger/service/S3LogUploaderServiceTest.java
RVelp1: ok thanks , i will do that

cerberus-audit-logger-athena/src/main/java/com/nike/cerberus/audit/logger/listener/AthenaLoggingEventListener.java
shawn-sher: It's easier to read

cerberus-auth-connector-okta/src/test/java/com/nike/cerberus/auth/connector/okta/OktaApiClientHelperTest.java
RVelp1: removed this test

cerberus-audit-logger-athena/src/test/java/com/nike/cerberus/audit/logger/service/AthenaServiceTest.java
RVelp1: made changes

cerberus-audit-logger-athena/src/test/java/com/nike/cerberus/audit/logger/service/S3LogUploaderServiceTest.java
RVelp1: made changes

cerberus-audit-logger-athena/src/test/java/com/nike/cerberus/audit/logger/service/S3LogUploaderServiceTest.java
RVelp1: made changes

shawn-sher (approved):

cerberus-audit-logger-athena/src/test/java/com/nike/cerberus/audit/logger/AthenaClientFactoryTest.java
```diff
@@ -0,0 +1,16 @@
...
+import com.amazonaws.services.athena.AmazonAthena;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class AthenaClientFactoryTest {
+
+  @Test
+  public void testGetClientAlwaysReturnsSameAthenaInstance() {
+    AthenaClientFactory athenaClientFactory = new AthenaClientFactory();
+    AmazonAthena clientInstance1 = athenaClientFactory.getClient("region-2");
```
melanahammel, line 12: What happens if this region is null or an empty string? Can you add tests for if `getClient` is called with bad parameters?

cerberus-audit-logger-athena/src/test/java/com/nike/cerberus/audit/logger/S3ClientFactoryTest.java
```diff
@@ -0,0 +1,16 @@
...
+import com.amazonaws.services.s3.AmazonS3;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class S3ClientFactoryTest {
+
+  @Test
+  public void testS3ClientFactoryAlwaysReturnSameInstance() {
+    S3ClientFactory s3ClientFactory = new S3ClientFactory();
+    AmazonS3 s3Instance1 = s3ClientFactory.getClient("region-1");
```
melanahammel, line 12: See comment above, what happens if `getClient` is called with a null or empty string?

cerberus-audit-logger-athena/src/test/java/com/nike/cerberus/audit/logger/service/S3LogUploaderServiceTest.java
melanahammel: This test is still here, can you uncomment or remove this test?

cerberus-audit-logger-athena/src/test/java/com/nike/cerberus/audit/logger/service/S3LogUploaderServiceTest.java
melanahammel: What changes were made, this test is still saying `assertTrue(false)`?

RVelp1: I think i have changes in my local, i will take a look and verify that

cerberus-audit-logger-athena/src/test/java/com/nike/cerberus/audit/logger/listener/AthenaLoggingEventListenerTest.java
```diff
@@ -0,0 +1,318 @@
...
+    Mockito.when(auditableEventContext.getOriginatingClass()).thenReturn("originatingClass");
+    Mockito.when(auditableEventContext.getPrincipalName()).thenReturn("pricinpleName");
+    Mockito.when(auditableEvent.getAuditableEventContext()).thenReturn(auditableEventContext);
+    Mockito.when(auditableEventContext.getPrincipalAsCerberusPrincipal())
+        .thenReturn(Optional.empty());
+    athenaLoggingEventListener.onApplicationEvent(auditableEvent);
+    Mockito.verify(auditLogger).info(Mockito.anyString());
+  }
+
+  @Test(expected = NullPointerException.class)
```
melanahammel, line 97: All of these tests are testing if certain items are missing in context. Can we discuss what this test and the ones below are testing? It looks like you are just adding more information to the auditableEventContext, but necessarily testing functionality.

cerberus-audit-logger-athena/src/test/java/com/nike/cerberus/audit/logger/AthenaClientFactoryTest.java
RVelp1: Test
the test is failing as shown in the image , i dont know what the alphanumeric  it's not matching with expected even i passed empty string at both places. Please take a look at this pic

RVelp1: Code is not throwing an exception for bad parameters.

cerberus-audit-logger-athena/src/test/java/com/nike/cerberus/audit/logger/S3ClientFactoryTest.java
RVelp1: image
same scenario here , am i doing anything wrong ???

cerberus-audit-logger-athena/src/test/java/com/nike/cerberus/audit/logger/listener/AthenaLoggingEventListenerTest.java
RVelp1: I'm trying to test insertion null values for each key that we are putting in Immutable hashmap and expecting NullPointerException having property in the message , but when comes to event_timestamp without checking null we are trying to format it so we are getting NullPointerException, so could not test this  like the way we are
testing other properties.

melanahammel (approved):

PR #562, opened 2021-06-24
Removed SpinnakerCodeStylePlugin

shawn-sher:
Removing the Spinnaker style plugin for Spotless since it's only published in bintray as far as I can tell.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,7 +16,6 @@
 
 
 import com.github.spotbugs.snom.SpotBugsReport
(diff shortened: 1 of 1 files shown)


slichlyter12 (commented):
```java
maven {
      url "https://dl.bintray.com/spinnaker/gradle"
    }
```

I think we can delete these lines (37-39), I think this plugin is the only thing coming from there.

slichlyter12 (approved):

